PR #2, opened 2019-10-23
Added phi^4 action in new script

c-j-johnston:


phi_four.py
```diff
@@ -0,0 +1,233 @@
...
+            else:
+                rejected += 1
+    print('Accepted: '+str(accepted)+', Rejected:'+str(rejected))
+    print('Fraction accepted: '+str(accepted/(accepted+rejected)))
+    return sample_distribution
+
+def main():
+    # set seed, hopefully result is reproducible
+    torch.manual_seed(0)
+    a_left, b_right = split_transformation(L)
```
wilsonmr: I'm pretty sure we don't want to use these. I had them in the other training example because my action was defined for a flattened state, where as the model returns a flattened split state i.e (even sites, odd sites). Since your action is defined for split states then the interaction pattern is not nearest neighbour geometry but something slightly different.

I think since you use this transformation everywhere then it technically won't matter if everything is self consistent although we should still sort this out so it's obvious what we're doing

phi_four.py
```diff
@@ -0,0 +1,233 @@
...
+
+import numpy as np
+import matplotlib.pyplot as plt
+
+import torch
+import torch.optim as optim
+
+from norm_flow_pytorch import NormalisingFlow, shifted_kl
+
+class array_mapping(object):
```
wilsonmr: I agree this can be slimmed down. We can discuss tomorrow but I think a single function is sufficient. I wrote a proposal one based on your code and checked they do the same thing

phi_four.py
```diff
@@ -0,0 +1,233 @@
...
+phi = phi_obj.flattened
+indices_array = phi_obj.indices_array
+shift = phi_obj.nn_arrays
+phi = torch.from_numpy(phi)
+# A_NP = np.array(np.random.rand(N_UNITS, N_UNITS), dtype=np.float32)
+# A = torch.from_numpy(A_NP)
+# COV_TARGET = A@A.T
+# INV = torch.inverse(COV_TARGET)
+
+def target_distribution_s(phi, m_sq, l, shift):
```
wilsonmr: I played with making this a subclass of `nn.Module` because then we can use torch.jit.trace which performs a JIT 'just in time' compilation of the forward function. It turns out this doesn't affect peformance however I would argue that the class is a neater way to arrange this.

Again we can speak tomorrow about this

wilsonmr:
Overall looks pretty good however, and since the metro appears to be getting a decent amount of acceptance it sounds like it's working to some extent!

wilsonmr:
@c-j-johnston @marshrossney Just walked past Luigi and I think evaluating the condition with logs is gonna be fine so I think the expression is

```
min(0, log_ptilde[i] + S[i] - log_ptilde[j] - S[j])
```

but I just wrote that down yesterday. Hopefully this is less likely to produce `nan`s in first place although I think we should also check for these as discussed

Here is the suggested code for getting shift and action. Take a look and see if you understand/agree with it and feel free to change variable names - I'm not super creative

https://github.com/wilsonmr/norm_flow_model/commit/e77a9085358c263b35b3f1b3a451dc1e6a868f8c

jmarshrossney:
Would have to draw random numbers from a log distribution this way. I'm not 100% sure, but perhaps it could be more efficient to do something like:
```
exponent = min(0, log_ptilde[i] + S[i] - log_ptilde[j] - S[j])
P_accept = np.exp(exponent)
```

If so, might be easier to return simple_prob_z_out*jacob_contr from NormalisingFlow.forward() and take the log within shifted_kl()

wilsonmr:
> Would have to draw random numbers from a log distribution this way. I'm not 100% sure, but perhaps it could be more efficient to do something like:
> `exponent = min(0, log_ptilde[i] + S[i] - log_ptilde[j] - S[j]) P_accept = np.exp(exponent)`

Yeah sorry I really meant just the conditional had the log. I agree with exponentiating after. I think we'll be exponentiating a more reasonable number here (hopefully)

> 
> If so, might be easier to return simple_prob_z_out*jacob_contr from NormalisingFlow.forward() and take the log within shifted_kl()

c-j-johnston:
I've changed the code so it calculates the exponent and only exponentiates `min(0, exponent)`. A few things I noticed: 
> `exponent = min(0, log_ptilde[i] + S[i] - log_ptilde[j] - S[j])`

isn't quite right; the signs of the S[i] and S[j] should be swapped here. 
I checked and I wasn't getting `nan`s or `inf`s before switching to using exponents, but I've switched anyway because it's faster.
In terms of speed, it cuts off around 0.3s (for a target chain length of 10000) if we use `np.exp(float(exponent))` instead of `np.exp(exponent)`. It's also around 0.2 s faster using `min()` rather than `np.min()` or `torch.min`.

wilsonmr:
> I've changed the code so it calculates the exponent and only exponentiates `min(0, exponent)`. A few things I noticed:

very good! When you get a chance take a look at the code I linked wrt action and shift calculation

> 
> > `exponent = min(0, log_ptilde[i] + S[i] - log_ptilde[j] - S[j])`
> 
> isn't quite right; the signs of the S[i] and S[j] should be swapped here.

I'm confused, I start with eq 6

eq1

but then use the notation of your code where \phi' is phi[j] and \phi^{i-1} is \phi[i]

eq2

I then sub in def of p

eq3

noting the minus sign infront of the action and then get

eq4

which is what I got, but now looking at your code you have absorbed the minus sign into S, which I personally find slightly confusing however in the end we are calculating the same thing.

> I checked and I wasn't getting `nan`s or `inf`s before switching to using exponents, but I've switched anyway because it's faster.

Yes I think this was just because the propsed phi were so uncharacteristic when using random numbers, the fact that you aren't getting `nan`s **and** that your acceptance is good really is a nice sign that we are doing something right!!

> In terms of speed, it cuts off around 0.3s (for a target chain length of 10000) if we use `np.exp(float(exponent))` instead of `np.exp(exponent)`. It's also around 0.2 s faster using `min()` rather than `np.min()` or `torch.min`.

this probably isn't the biggest time save in our entire pipeline but nice spot. I would argue that if it is at the cost of potential bugs in the case of us getting `nan`s then I'd rather take the extra few seconds (being bitten by this sort of thing before has made me cautious)

phi_four.py
```diff
@@ -0,0 +1,237 @@
...
+        i = np.random.randint(n_large) # random index to start sampling from configurations tensor
+        used = [] # track which configurations have been added to the chain
+        used.append(i) 
+        while chain_len < target_length:
+            j = np.random.randint(n_large) # random initial phi^j for update proposal
+            while j in used: # make sure we don't pick a phi we have already used
+                j = np.random.randint(n_large)
+            # ratio = (p_tilde[i]/p_tilde[j]) * (p[j]/p[i]) # calculate the ratio in A
+            exponent = log_ptilde[i] + S[j] - log_ptilde[j] - S[i]
+            P_accept = np.exp(float(exponent)) # much faster if you tell it to use a float
```
wilsonmr: I would imagine this is because you have a torch tensor (exponent) which numpy has to work the typing out for. Possibly, but maybe not `exponent.exp().item()` would be faster or the same as what you're doing

although they are similar remember numpy != pytorch and so using numpy functions on pytorch objects means that the object is probably copied along the way.. (more memory more time etc.)

wilsonmr: nope yours is faster! 👍 
```python
In [1]: import torch                                                            

In [2]: import numpy as np                                                      

In [3]: a = torch.rand(1)                                                       

In [4]: a                                                                       
Out[4]: tensor([0.2411])

In [5]: %timeit min(1, a.exp().item())                                          
2.6 µs ± 21 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [6]: %timeit min(1, np.exp(float(a)))                                        
2.09 µs ± 18.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

In [7]: %timeit min(1, float(a.exp().item()))                                   
2.72 µs ± 36.3 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

wilsonmr: mind you...
```python
In [8]: from math import exp                                                    

In [9]: %timeit min(1, exp(float(a)))                                           
596 ns ± 3.19 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

c-j-johnston:
> I then sub in def of p
> 
> eq3
> 
> noting the minus sign infront of the action and then get
> 
> eq4
> 
> which is what I got, but now looking at your code you have absorbed the minus sign into S, which I personally find slightly confusing however in the end we are calculating the same thing.

I think you're right about where I absorbed the minus sign being confusing so I've changed that. Meant to say I've started a new local branch incorporating the new action/shift - I'll push this later. Although I'm starting to think maybe I should have just done this on the branch I was on...

phi_four.py
wilsonmr: this all said, we should worry about optimising later

wilsonmr:
quite possibly, is the new branch based on this branch? In other words did you do something like
```bash
git checkout <thisbranch>
git checkout -b <newbranch>
```

wilsonmr:
because if so it's fine we can just merge the two on here with PRs!

c-j-johnston:
Yes - still getting to grips with collaborating on git so not the best at it yet, but I will create a pull request!

jmarshrossney:
Is there any advantage in continuing to use torch after the training - i.e. in the MCMC, post processing, plotting stages? We could just convert everything to numpy to avoid confusion and accidental mixing of the two.

wilsonmr:
i mean both the action and the model recieve tensors and should return tensors. Now since we get log_ptilde and S at start of chain if you really want you can do conversion there but I really think there are better things to focus on right now then over optimising something which takes seconds, when the training takes minutes and we're not in 'production' mode

jmarshrossney:
fair point

wilsonmr (approved):
There are some stylistic things but I think it's best if we get this on master (it looks like it's doing the correct thing) and then people can open smaller PRs if they want to, for example, improve MCMC or whatever

wilsonmr:
is  models/test_train.pt a valid model? Honestly I don't think the github page is the right place to store this, perhaps we can add models in the future as like final results.

If you are sentimental about it then run `git rm models/test_train.pt --cached` to keep a local copy but take it off the git repo and then commit that change and push, then we can merge

c-j-johnston:
Sounds good, but I'm now getting an error I wasn't before:

`AttributeError: 'NormalisingFlow' object has no attribute 'size_in'`

c-j-johnston:
And no, I don't think that model is valid any more. The 4x4 ones were trained with `m_sq=lambda=1` because they weren't given in the paper for that size. I'll label models better from now on but that one can go.

wilsonmr:
> Sounds good, but I'm now getting an error I wasn't before:
> 
> `AttributeError: 'NormalisingFlow' object has no attribute 'size_in'`

That's very weird because if you see in the diff, this PR is adding it

wilsonmr:
wait was that you loading the model? If so then the class has changed (again see diff)

c-j-johnston:
Yeah, I was using a locally saved model - I imagine it would work if I train a new one. I'll let you know once I've retrained

wilsonmr:
Yeah sorry all your saved models will be broken by this. I think it's better style though, my fault for not adding it before tbh

This could happen again so I would err on the side of not investing too much computing time into a model which may later become deprecated - hopefully things should start getting more stable soon

c-j-johnston:
It's working again now I've trained a new model. Just trying to keep a couple of models so I can check things are working without training every time. One thing I noticed - probably not something to focus on too much for now - is that acceptance for 6x6 is about half that of 4x4. I imagine the model is just under-trained at this size but it's quite a large drop-off in acceptance which I guess would only get worse for larger systems so we'll have to be wary of that in future.

wilsonmr:
yeah well they speak about training their models for weeks on GPUs so if we can get 30% in 10 minutes on a laptop I'm not overly concerned 😆 

Ok I will delete the deprecated example and then merge this, great job.

PR #3, opened 2019-10-23
use tensor indexing instead of sparse matrix masks

### LINKED ISSUE #1: get rid of mask matrices and instead use `torch.index_select`
as above I should check if I can get rid of the matrix transformations and use torch.index_select instead, probably it's more efficient and removes multiplication by incredibly sparse tensors

wilsonmr:
closes #1 

The overall functionality of the code shouldn't have changed. Might see a performance increase for larger systems

I have no idea how efficient `torch.cat` is but anyway this seems more idiomatic than the way I did it before and requires less explanation


wilsonmr:
appears to give minor performance increase and reduce complexity so I'm going to merge

PR #4, opened 2019-10-25
added new action and shift methods

c-j-johnston:
Changed action and shift methods


wilsonmr (approved):
I removed a bunch of globals because they were things we actually wanted to change and just seemed like bad coding practise in this instance. If we are to have globals I'd appreciate it if they were named in UPPER_CASE style

Other things were small fixes mostly convention like imports being standards, third party and locals in that order

otherwise looks good

PR #7, opened 2019-10-28
Stop getting NaNs for larger lattice sizes

### LINKED ISSUE #6: nans and infs for L>=8
I've been trying to train for L to get an idea of training times/potential optimisations but the loss values for lattices of 8x8 and above all give `-inf` and `nan`. Looking more closely, we get an array of `nan` when we print the output of line 98 of `phi_four.py`:  
`phi = model.inverse_map(z)`.
I've gone further into the code and checked the lines in `norm_flow_pytorch.AffineLayer.inverse_coupling_layer`. All are fully defined when calculating the first loss, but this loss still comes to `nan`. Then when computing the second loss, `s_out` and `t_out` and therefore `phi_b` and the left half of `phi_out` are arrays of `nan`s -- this makes me think the problem could be in the `_s_output` and `_t_output` methods; possibly something to do with the activation functions.
I've tried increasing the number of affine layers and their sizes in the model to no avail. L=6 is still working fine.

wilsonmr:
Realised that the way log p tilde was calculated was unneccessarily exponentiating, multiplying and then taking log of things. Fixed this by calculating in more sensible way, appears to have removed `nan` issue for now

closes #6

### DIFF
--- a/norm_flow_pytorch.py
+++ b/norm_flow_pytorch.py
@@ -9,7 +9,7 @@ distribution as per https://arxiv.org/abs/1904.12072
 Key imports from this module are NormalisingFlow and shifted_kl. Which are the
 necessary ingredients to train your own normalising flow model.
 """
-from math import sqrt, pi
+from math import sqrt, pi, log
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


norm_flow_pytorch.py
```diff
@@ -298,6 +267,7 @@ def __init__(
         ):
         super(NormalisingFlow, self).__init__()
         self.size_in = size_in
+        self._gaussian_normalization = log(sqrt(pow(2*pi, size_in)))
```
wilsonmr: probably makes little difference but we don't need to calculate this everytime so I made the log_probability_norm a function of the class so we can cache it

wilsonmr:
Do you agree that this fixes the issue? I briefly tried L=16 and it seems better

c-j-johnston (commented):
The fix is working for me - thanks!

norm_flow_pytorch.py
```diff
@@ -350,10 +340,10 @@ def forward(self, phi_input: torch.Tensor) -> torch.Tensor:
         jacob_contr = torch.ones(phi_input.size()[0], 1)
```
wilsonmr: oh dear this should be zeros since it's actually log of jacobian_contribution!

PR #11, opened 2019-10-29
Observables

jmarshrossney:
More important: observables.py. Nothing too complicated going on but plenty of room for mistakes.

Not particularly important: just split phi_four.py into train.py and sample.py. Also made separate modules for indexing functions (grid.py) and global parameters (params.py).


wilsonmr:
small comments:
 - try and sort out the conflict, I think the best will be to rebase this on master, and just let `phi_four.py` not be deleted and then throw a new commit in where it gets deleted or something. The longer we leave this I think it'll get more annoying to sort out
 - I quite like the modular system, it's not too far from being sorted however doing things like `from params import *` is generally considered a bad idea, as discussed today I think we should look at parsing these from a yaml/json file. But I have quite a clear idea of how I want to do this and so I'm happy to use params.py as a temporary solution and focus on other things first.
 - there are quite a few whitespace issues but these are largely cosmetic. I personally use VS code to edit python and find the stylistic comments (from pylint) on the whole make my code look a lot nicer, but up to you how you want to code ofc.

I will try and look in more detail this week

jmarshrossney:
Cheers. Should be able to merge now.

Please do continue to flag up bad practice and stylistic issues; I'm genuinely grateful for being made aware of them.

PR #12, opened 2019-10-30
Remove backward shifts

gfilaci:
Thanks to translation invariance, it is possible to rewrite the action in a standard form with forward (or backward) shifts only.

### DIFF
--- a/phi_four.py
+++ b/phi_four.py
@@ -17,10 +17,45 @@ import torch.nn as nn
 from norm_flow_pytorch import NormalisingFlow, shifted_kl
 
 def get_shift(length: int) -> torch.Tensor:
-    r"""Given a 2D state of size lengthxlength returns a 2x(length^2)
-    tensor where each row gives the 2 nearest neighbours to a flattened state
-    which has been split into (\phi_even, \phi_odd) where even/odd refer to
-    parity of the site
+    r"""Given length, which refers to size of a 2D state (length * length)
+    returns a 2x(length^2) tensor where each row gives the 2 nearest neighbours
+    to a flattened state which has been split into (\phi_even, \phi_odd) where
+    even/odd refer to parity of the site.
+
+    Parameters
+    ----------
+    length: int
+        Defines size of 2D state (length * length)
+    Returns
+    -------
+    shift: torch.Tensor
+        Tensor which can be used to index flattened, split states such that
+
+            state = tensor([\phi_even, \phi_odd]),
+
+        then state[shift] will return a 2xlength tensor:
+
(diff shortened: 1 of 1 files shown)


wilsonmr:
a very good point!!

wilsonmr:
@c-j-johnston take a look at this, if you're happy then please merge

phi_four.py
```diff
@@ -75,7 +73,7 @@ def forward(self, phi_state: torch.Tensor) -> torch.Tensor:
         action = (
             (2+0.5*self.m_sq)*phi_state**2 + # phi^2 terms
             self.lam*phi_state**4 - #phi^4 term
```
tgiani, line 149: @c-j-johnston  shouldn 't you multiply times 0.5 here to be consistent with the other terms of the action? And also, everything is multiplied by a factor of 0.5 with respect to the action of Eq. 20 in the paper. Is this correct?

gfilaci: I've tried to implement the optimisation
\sum_x phi(x) ( phi(x-mu) + phi(x+mu) ) = 2 \sum_x phi(x) phi(x+mu)
where I've shifted x -> x+mu in the first term.
The idea is that if you are summing pairs of neighbours over the whole lattice, the order doesn't matter.
In other words, when you plug eq (21) into eq (20) you see that the two terms with shifts in eq (21) give actually the same contribution to the action.
The action in (20) is unchanged: the difference is that instead of doing 0.5*(shift_forward_term + shift_backward_term) we can just do simply 0.5\*2\*shift_forward_term=shift_forward_term.
I tried a minimal modification, hoping not to spoil the shift functions... but I don't exclude that in the code there are other places where having backward and forward shifts is assumed.
You can also try to compute the action of a configuration with both formulations and check that the results are the same (it works for me for L=2,4).

gfilaci: The other factors look ok to me (but maybe it is easier to check having our action in latex...)

wilsonmr: Added some notes on this and improved documentation after our discussion

wilsonmr:
I'm really happy for this to be merged, I was hoping Cameron would get a chance to look at it, just to make sure he understood the change but I guess we can just discuss at next meeting, gonna merge

c-j-johnston:
Apologies, I've only just had the chance to look - is the idea that every nearest neighbour pair is counted twice if we sum over the whole lattice, so instead we are only using two of the four directions and including a factor of 2?

gfilaci:
Yes exactly, in this case it's enough to consider the two independent directions with positive displacement (clearly this might change when we'll look at CP1)

PR #13, opened 2019-11-04
Training metrics

jmarshrossney:
This seems like a significant enough addition to justify a new branch, though it's based on the "observables" branch. The plan is to rebase this branch on observables each time something is changed.

The main idea is to use the acceptance rate and autocorrelation time as training metrics alongside the loss, which means running the sampling step at intervals throughout the training. This is useful because we can see if, say, the loss is continuing to decrease but we're getting negligible improvements in acceptance, or vice versa.

There's also some very basic plotting functionality which is useful in that it allows easy comparison of similar training runs, which differ by perhaps one or two parameters.


jmarshrossney:
I will re-write this so that it fits with the more recent "observables_fix" branch then make another pull request.

Issue #9, opened 2019-10-28
Momentum space two-point function is not convex


jmarshrossney:
This is to flag up a problem I'm looking into at the moment.

The 0-momentum two-point function should be convex (see Fig 3 in [this paper](https://arxiv.org/pdf/1904.12072.pdf)), but currently this is not the case.

This is likely just a mistake in my code somewhere. I'm looking into it, but thought I should make everyone aware. The code is in `observables.py`.

There is also the possibility that it could be due to under-training.

Note that the effective pole mass formula is nonsense when $\tilde G_c (0,t)$ is not convex.

jmarshrossney:
Resolved in #14

PR #14, opened 2019-11-04
Minimally add observables

wilsonmr:
I tried to implement observables from #11 without having as much restructure because it made it difficult to focus on the important thing which was the actual observables implemented.

In particular I felt like `get_shift` just needed to be slightly extended to be useful for both action and observables

I think there was a mistake with the two point green function which I fixed, I also leverage tensor indexing to do this in a way which I hope is easier to read

I trained a 6x6 which ends up with 45% acceptance, which I think is at least sufficient to test this. But the observables I'm getting are not matching the paper. This could be a for a number of reasons.

### DIFF
--- a/geometry.py
+++ b/geometry.py
@@ -15,9 +15,12 @@ def get_shift(
         dims: tuple = (0, 1),
         ) -> torch.Tensor:
     r"""Given length, which refers to size of a 2D state (length * length)
-    returns a 2x(length^2) tensor where each row gives the 2 nearest neighbours
-    to a flattened state which has been split into (\phi_even, \phi_odd) where
-    even/odd refer to parity of the site.
+    returns a Nx(length^2) tensor where N is the length of `shifts` and `dims`
+    (which must be equal). Each row of the returned tensor indexes a flattened
+    split state \phi = (\phi_even, \phi_odd) which is split according to a
+    checkerboard geometry (even and odd refer to parity of the site). The
+    indices refer to shifts on the states in their original 2D form. By default
+    N = 2 and get_shift simply returns the right and down nearest neighbours.
 
     Parameters
     ----------
@@ -26,13 +29,15 @@ def get_shift(
     shifts: tuple
         a tuple of shifts to be applied. Each element represents a shift and can
         either be an integer (if the shift is in a single dimension) or a tuple
-        if the shift is applied simultaneously in multiple dimensions.
-        By default it is set to have two shifts, left 1 and up 1 (which in turn
-        allow for right and down nearest neighbours)
+        if the shift is applied simultaneously in multiple dimensions (see
+        Examples). By default it is set to have two shifts which give right and
+        down nearest neighbours
     dims: tuple
         a tuple of dimensions to apply `shifts` to. As with shift, each element
         in dim can itself be a tuple which indicates that multiple shifts will
-        be applied in multiple dimensions simultaneously.
+        be applied in multiple dimensions simultaneously. Note that
+        corresponding entries of dims and shifts must also match (either both
+        ints or both tuples of same length).
     Returns
     -------
     shift: torch.Tensor
@@ -61,7 +66,8 @@ def get_shift(
     tensor([[1, 2, 0, 3],
             [2, 1, 3, 0]])
 
-    correct nearest neighbours (left and down) are given in each row respectively
+    correct nearest neighbours in reference to the original `state_2d` (left and
+    down) are given in each row respectively
 
     to see how multiple shifts works, consider the shift (1, 1)
     (left one, up one)
@@ -73,15 +79,25 @@ def get_shift(
     we see that each element of shifts and dims can perform multiple shifts, in
     different dimensions at once.
 
+    Notes
+    -----
+    The conventions for how the shifts are applied are according the torch.roll
+    function, shift = +ve rolls the state left and so the indices will refer to
+    lattice sights to the right.
+
+    See Also
+    --------
+    torch.roll: https://pytorch.org/docs/stable/torch.html#torch.roll
+
     """
     if len(shifts) != len(dims):
         raise ShiftsMismatchError(
             "Number of shifts and number of dimensions: "
-            f"{len(shifts)} and {len(shifts)} do not match."
+            f"{len(shifts)} and {len(dims)} do not match."
         )
-    # define a checkerboard
+    # TODO: make a Geometry2D class which is initialised with a matrix
+    # (like checker) which tells you the split geometry of the state
     checkerboard = torch.zeros((length, length), dtype=torch.int)
-    # set even sites to 1
     checkerboard[1::2, 1::2] = 1
     checkerboard[::2, ::2] = 1
 
--- a/geometry.py
+++ b/geometry.py
@@ -1,0 +1,111 @@
+"""
+geometry.py
+
+Module containing transformations related to geometry.
+"""
+from math import ceil
+
+import torch
+
+class ShiftsMismatchError(Exception): pass
+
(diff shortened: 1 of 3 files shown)


observables.py
```diff
@@ -0,0 +1,90 @@
...
+    m_t = []
+    for i, g_0_t in enumerate(g_0_ts[1:-1], start=1):
+        m_t.append(acosh((g_0_ts[i-1] + g_0_ts[i+1])/(2*g_0_t)))
+    return m_t
+
+def susceptibility(g_0_ts):
+    """Calculate the susceptibility, which is the sum of two point connected
+    green functions over all seperations. Since we are calculating the zero
+    momentum green function we can just sum over t"""
+    return sum(g_0_ts)
```
wilsonmr: @tgiani do you agree with this? I don't know if I am saying something stupid but if we get zero momentum GF and sum on t, I think this should be correct?

wilsonmr:
@tgiani also would be good to test with L=2, did you make progress on the notebook?

phi_four.py
```diff
@@ -181,82 +107,119 @@ def train(model, action, epochs):
...
+        if ~np.isfinite(exp(float(min(log_ratio) - max(log_ratio)))):
+            raise ValueError("could run into nans")
+
+        if not batches: # first batch
+            log_ratio_i = log_ratio[0]
+            start = 1
+        else:
+            start = 0 # keep last log_ratio_i and start from 0
+
+        for j in range(start, n_large):
```
wilsonmr, line 169: I think there was no need to have the slightly complex random phi, instead can just work down the generate sample in order. This then allows us to easily 'batch' the sampling for larger samples. Now I can get approx 1e6 samples in 100 seconds

geometry.py
```diff
@@ -0,0 +1,111 @@
...
+        in dim can itself be a tuple which indicates that multiple shifts will
+        be applied in multiple dimensions simultaneously.
+    Returns
+    -------
+    shift: torch.Tensor
+        Tensor which can be used to index flattened, split states such that
+
+            state = tensor([\phi_even, \phi_odd]),
+
+        then state[shift] will return a 2xlength tensor:
```
wilsonmr, line 48: this needs updating

observables.py
tgiani: uhm do you want to implement Eq.(25) of the paper? I m not sure.. Who is ```length``` in ``` zero_momentum_green_function```? It should be the full size of the lattice, right?

tgiani:
> @tgiani also would be good to test with L=2, did you make progress on the notebook?

Yep we can have a look tomorrow

observables.py
tgiani: ah no right, sorry got it now

tgiani: I think is correct

wilsonmr: If I take eq. 23 and set p=0 and then take sum_t the question is do I get RHS of eq. 25

the length here is just used for get_shift not really used in the calculation

tgiani: To me it looks correct.. You have to sum over all the points ```x``` of the lattice and you can do this  by summing first over the spatial component and then over the time one

wilsonmr: ok good, then we should understand why I can get acceptance of like 45% but the numbers are way off

tgiani: Do you have numbers for L=2?

wilsonmr: no this is L=6, I can get L=2 pretty quickly later

wilsonmr: but what should I set M^2 and lambda to?

tgiani: just set them to 1, I ll use the same in the notebook

wilsonmr: ok with these I get:
```
Accepted: 100000, Rejected: 11351, Fraction: 0.9
Ising Energy: 0.057270217686891556
Susceptibility: 0.3175181970000267
```

tgiani: I m getting susceptibility ```0.348916```

tgiani: I don t have ising energy yet

wilsonmr: ok if I do more training and then add a random interval between samples when calculating the observables of 5 to account for some autocorrelation then I get
```
Accepted: 100000, Rejected: 6044, Fraction: 0.94
Ising Energy: 0.0637047290802002
Susceptibility: 0.3435881659388542
Time to run MC for a chain of 100000 samples on an L=2 lattice: 4.277069807052612 seconds
```
I think this looks like the training, sampling and susceptibility are at least ok

geometry.py
```diff
@@ -0,0 +1,111 @@
...
+    tensor([[3, 0, 2, 1]])
+
+    we see that each element of shifts and dims can perform multiple shifts, in
+    different dimensions at once.
+
+    """
+    if len(shifts) != len(dims):
+        raise ShiftsMismatchError(
+            "Number of shifts and number of dimensions: "
+            f"{len(shifts)} and {len(shifts)} do not match."
```
jmarshrossney: one of the `len(shifts)` should be `len(dims)`. This is a potentially catastrophic bug.

observables.py
jmarshrossney: In the most recent version of the [https://arxiv.org/pdf/1904.12072.pdf](paper) they corrected a mistake in equation 23, meaning we should be taking the mean, not sum, over `x` for each `g_0_t` in the zero momentum Green's function (and, as a consequence, the `susceptibility` function will have to be altered).

The agreement for L=2 susceptibility is really encouraging!

geometry.py
wilsonmr: > one of the `len(shifts)` should be `len(dims)`.

Good spot, will correct

> This is a potentially catastrophic bug.

ah not really, the error gets raised correctly, it's just the error message will be pretty useless, which completely defeats the point of making its own class with a sensible name it would be catastrophic if it didn't get raised

observables.py
wilsonmr: the action is defined incorrectly in the arxiv version but the momentum green function is incorrect in the DOI.. So annoying 😆

jmarshrossney (approved):

geometry.py
jmarshrossney: Fair enough, irony probably doesn't belong on Github review comments.

jmarshrossney:
> I think there was a mistake with the two point green function which I fixed, I also leverage tensor indexing to do this in a way which I hope is easier to read

If it's not too much trouble, briefly, what mistake did you spot in the previous implementation of the two point function? The new shift function and tensor indexing makes it far clearer, I agree, but I can't seem to see any major difference in the new version other than that it doesn't average over displacements in all four possible directions for each site and 'x' (which I now see was unnecessary).

Fantastic that we're getting agreement between lattice and analytical now.

I think issue #9 can probably be resolved too, since fixing the aforementioned mistake appears to have fixed this.

wilsonmr:
>If it's not too much trouble, briefly, what mistake did you spot in the previous implementation of the two point function?

Actually I think I was just getting mixed up because you were looping over lattice sites so `*.mean()` implictly was over the 0th dimension so I think it might have been fine, sorry about that

The convex issue I think comes from the fact that you were trying to calculate effective pole mass for t in [0 to L-1] instead of [1 to L-2] (all except end points as in paper) otherwise I just was implementing what you already had, but with a bit less of the restructuring

observables.py
gfilaci: Just a double check: with my notebook I get the same number for susceptibility, and 0.0642849 for the Ising energy, so it looks good!

gfilaci (approved):

c-j-johnston (approved):

geometry.py
wilsonmr: lol I'm terrible at reading tone 🎣

PR #15, opened 2019-11-05
Write files

c-j-johnston:
Added a script to write text files for parameters; these files are given in command line for training and become name of model. Allows training to be run by writing a shell script with different parameter files.

### DIFF
--- a/grid.py
+++ b/grid.py
@@ -1,0 +1,93 @@
+"""
+grid.py
+
+Module containing functions related to the grid, and transformations between the original
+2-d grid and the 1-d chequerboard "split state" as required by the neural networks.
+
+The only module not to import params.py - currently it's imported *by* params.py for
+the phi4 action computation.
+"""
+import torch
+
+def make_checkerboard(L: int):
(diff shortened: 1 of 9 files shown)


jmarshrossney:
I agree that it's probably more flexible to read parameters from a file than to have to modify `params.py` every time. I think Mikey is quite keen to do this using json though.

Also I'm gonna need to redo a lot of the training_metrics branch to fit with master (my bad for trying to  squeeze a major restructure in with a new functionality branch). I could do that and integrate your additions in at the same time - maybe that's the easiest?

wilsonmr:
json is nice because it is natively supported by python, yaml is nice because it is slightly more forgiving if I remember correctly and you can use the actual python names for `None`

I am quite keen to do a rather major restructure of code, and have a pretty clear idea of what it should look like.. I will try to do this today

Probably it will make other branches a little redundant so I would say that in the mean time maybe don't invest too much time in a feature because merging that into the restructured code will be horrible. Is that ok? I'm sorry to have steam rolled a couple of both of your PRs and it's mainly my fault for being lazy setting up the project in the first place!

PR #16, opened 2019-11-05
Big restructure

wilsonmr:
Hopefully won't need to do something quite so drastic again


wilsonmr:
some notes:

 - [x] get train and sample out of phi_four
 - [x] make a geometry class for 2D lattice
 - [x] action uses geometry, should make moving to higher dims easier if we want (just make new geometry class)
 - [x] make a config.py which parses a runcard in a somewhat sensible manner
 - [x] checkpoints
 - [ ] phi_four main just takes runcard as positional perhaps specify output folder and debug mode? (later)

wilsonmr:
@marshrossney @c-j-johnston I have ran out of time today however the main bulk of the restructuring is here

Note that you will need to change the upstream of your git repo if you want to push/pull:

```
git remote set-url origin https://github.com/wilsonmr/normflow_lattice.git
```

Then you need to install the code into the same virtual env as whatever python you're using:

```
python -m pip install -e . # in root of repo
```

Note this is a development installation*

Tomorrow I should be able to finish this, you can take a look at the example runcard - this is the level of flexibility I want to have (some things could have default values) but having this all specified means that we can remember what we did

Probably these runcards are slightly more laborious to change than your others, although I think manually changing them isn't too bad

*which means the python files are symlinked into the virtual env - this means if you change them then you don't need to reinstall (perhaps unless you add a file)

wilsonmr:
@gfilaci and @tgiani also see above if you were planning on doing anything with the code

wilsonmr:
ok I think this is working, I would like to add argparse to get the runcard and allow for command line args in a slightly better fashion but I think this roughly is correct

I took a few things out of the example runcard, we can add those features later.

@c-j-johnston @marshrossney could you check if you are able to install  and run something? (see above about changing the remote on your local git repo)

setup.py
```diff
@@ -0,0 +1,22 @@
+from setuptools import setup, find_packages
+
+with open('README.md') as f:
+    LONG_DESC = f.read()
+
+setup(
+    name="normflow",
+    version='0.1',
+    description="Normalising flow model on the lattice",
+    author="Michael Wilson",
```
wilsonmr, line 10: should add to this, was just filling the blank

wilsonmr:
oh btw the README.md is wrong, you just need to run ~`normflowphifour <runcard>`~ once you have installed

EDIT: `normflowphifour <runcard> <output directory name>`

wilsonmr:
@c-j-johnston I take the commits as answering yes to the above question of whether you can run something?

I appreciate the output folder thing, this was something I wanted to think about a bit more before merging this but after I was sure people could actually run/install this. I purposely left the exception in because I was half wondering about each training having a distinct output folder, into that output folder we would copy the specific runcard for that run and any other models/plots etc. This means that whenever you come across an output folder you can always see the configuration file which generated it. I admit that for retrains this is a bit annoying, but it does ensure nothing gets mixed up. For example atm if I want to do two trainings with slightly different hyper params and I forget to change the output file name then I will overwrite the first set of checkpoints, which might mean I need to rerun the training - potentially losing a weeks worth of computing time later on!

The training plot also looks nice, however I think it would be better if the `train` action saved a train log, with chi2s etc. and then this plot was not in the `train` action which should be a pure function which does model in -> model out

wilsonmr:
Also I would refrain from adding too much to this branch, as you can see from the diff it's already a nightmare and I pretty much just rearranged things and added some higher level configuration options. If we start adding new features then as well as checking I didn't break things by moving everything we also will need to check the new features are actually working

c-j-johnston:
Yes, apologies, I meant to say I had things working. With the outputs, I opened an isue (#17 ) suggesting don't specify a directory name and instead name the directory after the runcard, although I realise this means we would need to make sure runcards are given different names if they are edited. I'll try to plot the loss out of the function and create a log too.

wilsonmr:
if it's urgent then make `train` return an array which serves as a training log, later on this could include the acceptance/rejection frequency as well a la #13 

but then let's try and get this merged and then you can open small PRs which make all of this better (I was rushing to get the restructure done so we can all work in parallel more easily)

wilsonmr:
side note: I renamed the repo once but I honestly would really like to come up with a snappier name for this and then stick with it (since it's a pain everytime I rename the repo)... Ideally a shortish acronym which vaguely works

normflow_lattice is crap and too long. I threw some of the semantics into an acronym generator and it didn't come up with anything

since we're doing NVP I thought the name could include 'envy' like a phonetic version of NV (the p can be absorbed into the V or whatever) and then maybe something to do with the fact that this is specifically with lattice configurations. I'm less bothered if the name makes 100% sense and would rather have something short to type and sounds cool (priorities)

jmarshrossney:
This is OP as hell.

I'm having an issue right now where the memory use gradually increases until the job is automatically killed. I'm just running the example as
```
normflowphifour examples/runcards/phi_four_train.yaml test
```
Is this me being stupid or...?

The restructure is genuinely very cool though, and I'm sure it will streamline things later on.

wilsonmr:
> This is OP as hell.
> 
> I'm having an issue right now where the memory use gradually increases until the job is automatically killed. I'm just running the example as
> 
> ```
> normflowphifour examples/runcards/phi_four_train.yaml test
> ```
> 
> Is this me being stupid or...?

No, that sounds really bad! Is that in the training or sampling?

> 
> The restructure is genuinely very cool though, and I'm sure it will streamline things later on.

It's alright, I actually think I've over complicated some parts a bit and it would be simplified by having a train script and a sample script - then the runcard parsing could be a lot more simple and transparent

c-j-johnston:
I had a similar problem yesterday when I tried to run the code using SSH onto the school's `studentrun01` server, but didn't have any problems on my laptop. Remaining time for training was an expected time at first, then shot up to 24 hours and was killed.

jmarshrossney:
Ah it was `loss_list` getting huuuuge. I think a way to get round that would be to fix the length and save periodically.

wilsonmr:
Probably what we want is a `log_interval` which has a default value but can be overridden by runcard that saves to a log file in the output the loss every `log_interval` instead of keep the log as a list?

c-j-johnston:
Need to work on some other stuff first, but I'll do that this afternoon and commit it to this branch, unless it would be better just to remove the plot bit now and I'll add it in a new branch.

jmarshrossney:
> I purposely left the exception in because I was half wondering about each training having a distinct output folder, into that output folder we would copy the specific runcard for that run and any other models/plots etc. This means that whenever you come across an output folder you can always see the configuration file which generated it. I admit that for retrains this is a bit annoying, but it does ensure nothing gets mixed up. For example atm if I want to do two trainings with slightly different hyper params and I forget to change the output file name then I will overwrite the first set of checkpoints, which might mean I need to rerun the training - potentially losing a weeks worth of computing time later on!

I think this makes a lot of sense. Just thinking of that common situation where your data looks like it's doing something strange and you want to double check the input parameters. If there's no parameters file, or if you're not 100% certain that whatever parameters file is there corresponds to the data, things become un-fun.

wilsonmr:
> Need to work on some other stuff first, but I'll do that this afternoon and commit it to this branch, unless it would be better just to remove the plot bit now and I'll add it in a new branch.

hmm it's a tough one, I think keep it here, seems like a pretty useful thing from the get go

wilsonmr:
closing in favour of #20

PR #20, opened 2019-11-12
Reportengine implementation

wilsonmr:
Pretty much does the same as big restructure but also adding [reportengine](https://github.com/NNPDF/reportengine/) as a dependency

 - [x] fixed sampling
 - [x] made better training output
 - [x] added a few observables to get us started
 - [ ] update docstrings so that code is actually well documented and documentation is up to date

I would be interested if people can follow the installation instructions in the README and get this working

There are still lots of improvments to be made here, like adding logs, better specification of networks and more observables but this will at least get us started. In the end I think `reportengine` will be a useful tool for the analysis stuff and I do really like the way it's structured

## Notes on reportengine

basically reportengine adds the concept that our plots/tables are 'actions' and an action can depend on either another action (like MC sampling) or a runcard variable like `lattice_length` or a production rule (these are in `config.py` under the names `produce_*`) you define actions to take as arguments the dependencies, so for example

```
def ising_observables_table(ising_energy, susceptibility, training_output)
```

takes `ising_energy` and `susceptibility` (other actions) and `training_output` (specified in runcard) in turn `ising_energy` and `susceptibility` take `two_point_green_function` and that chain goes until the leaves of the tree are either set by default or specified from runcard. reportengine builds the dependency graph and works out all of the things you need to specify in the runcard and will complain at you until you get it right. For the function I mentioned the runcard keys required are `training_output` which points to a valid training output folder, `cp_id` which is the checkpoint which will be loaded and `target_length` which says how many samples we want.

Clearly in the future we will probably also want `thermalisation` and `sample_interval` but we can worry about that later

With this simple example reportengine doesn't look so special however we already can do the following report without any further coding:

```yaml
training_output: l2_train_example

# TrainingOutput object has a special `as_input()` attribute which allows for a special from_ key, in this case we take all checkpoints in the training output folder

cp_ids:
 from_: training_output

target_length: 100000

meta:
  author: New User
  title: Training Report
  keywords: [example]

# template text is markdown and then we can embed tables and figures by specifying actions inside special character {@ and @}
# loops over cp_ids each iteration of loop sets cp_id to a value using special {@with@} loop
template_text: |
 {@with cp_ids@} 
 ## {@cp_id@}
 {@ising_observables_table@}
 {@endwith@}

actions_:
  - report(main=True)
```

this is quite a lot to take in but I think once you get your head round it, it will make compiling results way easier. Try running the l2 example which takes a minute or so then running that report on it and see what happens.


wilsonmr:
as a side note, I can reproduce the L=6 results with 40% acceptance now that I fixed the sampling

jmarshrossney:
Working for me. I'm happy for this to become the master branch as soon as.

wilsonmr:
Ok nice! I think I will merge soon provided Cameron is also happy with this

c-j-johnston:
Apologies, I won't have time to check until tomorrow -- I don't mind if you just merge now though if it's working for everyone else!

wilsonmr:
I don't think we're in that much of a rush! Since yourself and Joe are probably the main users of this I'd rather make sure before merging. Once it is merged I will probably set each of you a very small coding task to try and get used to `reportengine` and the namespaces

jmarshrossney:
We're not in a huge rush, but the boot which will soon give me a kick up the arse is certainly already in motion. Luigi is expecting "good" results by Tuesday, i.e. ones with quantified uncertainties and taking autocorrelation and thermalisation into account.

So for me the priority would be to get started on those. If it makes more sense to work on these things in a separate branch and then make a pull request rather than add them here (which I think it does), then it would be nice if this was merged I guess.

Adding new plotting functions etc seems to be working for me.

jmarshrossney:
There is one thing I'm wondering about which is relevant to this pull request, which is output data files.

If you run `anvil-sample` with a runcard which, for example, makes a table of integrated autocorrelation times for different checkpoints, then you get a CSV file with that data.
But if you then decide you want to check the acceptance rate, you have to run a whole new sampling simulation, using a runcard which does both tables.
I can imagine finding myself using runcards which just do everything so I've got the data saved to a file, just to be safe.

You don't seem to get any output data with plots.
What if I just want to change a legend or something? Do I have to run `anvil-sample` again?

Would it make sense to save everything to output data files and plot from there, or would this be an unnatural use of reportengine? 
Or, does reportengine have any clever solutions for this?

wilsonmr:
So there are two things here:

As you point out the table saves a CSV file. Now if a table action is a dependency of a plotting action then the intermediate table action will still save the CSV file so although I haven't done this for the plotting action I implemented a good practise is to do:

```python
@table
def table_action(...):
    ...
    return dataframe

@figure
def plot_table(table_action):
    ...
    return mpl_figure_of_table
```

Then if you did wish to remake the plot without rerunning anything, you could do something like:

```python
>>> from anvil.observables import plot_table
>>> import pandas as pd
>>> df = pd.read_csv("filename", headers=[0], etc.) # need to properly define columns and indices here
>>> fig = plot_table(df)
>>> fig.<make edits on various things>
```

actually you don't even need to use the action here, you can just load the CSV file and make a new plot. Unfortunately that's about as deep as it goes.

The second thing is there is a way to save mpl figures in a binary format so they can be reloaded by mpl. I honestly don't know if reportengine supports this, but if not we could subclass figure and do it ourselves..

wilsonmr:
in terms of working on things, I'm hoping everything will be fine with this branch up to small changes, if it's just a case of adding additional functions then making a branch based on this one and rebasing when necessary is probably not gonna cause issues

wilsonmr:
Also if it wasn't apparent, I was planning on renaming the repo (for the final time - sorry!) to anvil upon merging this, and then I'll send a final email around for people to take a look at (and edit their upstream if necessary)

c-j-johnston (approved):
All working for me so I'm happy for it to be merged

wilsonmr:
ok thanks @c-j-johnston ! Well let's just go for it and if we find problems we can fix them with smaller PRs

PR #21, opened 2019-11-15
sample() returns decorrelated subsequence of configurations

jmarshrossney:
The `sample` function (in sample.py) now returns approximately* `target_length` configurations, which are separated in Monte Carlo time by `tskip`, which is at least double the integrated autocorrelation time of the Markov chain.

Added new runcard parameter: `tskip_guess`, which allows an instant estimate how many configurations to generate in order to obtain approximately `target_length` after subsampling.
If this is not specified, then the first batch of configurations will be used to provide a guess.

The next thing to do would be to check that observables show almost instant decorrelation.
If they do, the mean and variance, averaged over several separate simulations, should be reliable estimators.
I'll start working on these things straight away, but I thought it might be useful to make this pull request now, with just the one addition.

*if you give a wildly wrong `tskip_guess` then you won't get approximately `target_length` configurations.


c-j-johnston (approved):

anvil/sample.py
```diff
@@ -12,18 +12,132 @@
...
+    # Compute integrated autocorrelation
+    integrated_autocorrelation = 0.5 + torch.sum(
+        autocorrelations / (N - torch.arange(N - 1))
+    )
+
+    return integrated_autocorrelation
+
+
+def produce_samples(
+    loaded_model, action, length, state_i=None, log_ratio_i=None, first=False
```
wilsonmr: could we maybe call `length` something like `batch_size` I think it'll be more explicit

anvil/sample.py
```diff
@@ -12,18 +12,132 @@
...
+):
     r"""
     Sample using Metroplis-Hastings algorithm from a large number of phi
     configurations.
-
+
```
wilsonmr: trailing whitespace in general is frowned upon, at the very least it looks sloppy - fine... At the worst I think it can cause genuine problems

e.g https://softwareengineering.stackexchange.com/a/121560

running `pylint` on this file will make it complain, running `black` will just fix it for you (and do other formatting things according to a mixture of PEP and black specific rules

anvil/sample.py
```diff
@@ -34,64 +148,72 @@ def sample(loaded_model, action, target_length: int, n_large=20000) -> torch.Ten
...
-                accepted += 1
-                i = j
-            else:
-                accept_reject[j] = False
-                chain_indices[j] = i
-                rejected += 1
-        start = n_large*batch
-        end = start + batch_length
-        batches[start:end, :] = phi[chain_indices, :]
+    batch_length = min(target_length, 10000)  # hard coded for now
```
wilsonmr: fair enough, although I think we could have just kept `n_large=10000` in the function arguments, just incase somebody does have memory issues (might be an issue as we scale `lattice_length`)

wilsonmr: btw it can have a different name, maybe `batch_size` in general would work well - and make runcard configuration cleaner

wilsonmr:
see I half wonder if it wouldn't be cleaner to just have:
```
loaded_model, action, batch_size
|_ chain_autocorrelation
|  |
|__|_ sample
```

in which case we don't need to guess `tskip` it just works it out based on a single batch. Then you can still have a safety factor if you want a bit like the function. I think that estimating `chain_autocorrelation` will barely take any time compared to doing a decent sized sample

I would personally rename `produce_samples` to `sample_batch` to not confuse with production rules and then both `chain_autocorrelation` and `sample` will use `sample_batch` as a base level function rather than it being an action

anvil/sample.py
jmarshrossney: ach, I ran black and it's not changing it! Probably need to configure it or something..

wilsonmr: that's a shame, it's normally pretty good. I'm also running pylint on VS code so this kind of thing gets highlighted to me.

Sorry for being a bit pedantic but a wise person once told me a [story](https://www.insider.com/van-halen-brown-m-ms-contract-2016-9) of why if you see something like whitespace or no final newline in code them you instantly become suspicious that there could be other more serious

jmarshrossney:
Agree with suggestions for renaming variables and functions.

I think I agree with the suggested restructure, if I understand it correctly.
Yeah it's fast enough that we don't need to guess `tskip` ourselves.

Thinking about it, we don't need to use the entire chain to estimate the integrated autocorrelation time because
1. It almost never fluctuates enough to change the value of `ceil(2 * tau_int)`
2. This isn't at all guaranteed to lead to statistically independent values for observables anyway, so it can be a rough estimate. We should check the autocorrelation times for observables too.

I was struggling to understand their expression for the autocorrelation time for a while, because they claim it's observable independent, which doesn't square with the fact that the autocorrelation time for different observables scales differently with system size.
It's worth noting that their formula only holds for infinite chain length...

Anyway, we can use this as a first guess, check that observable autocorrelation functions are delta functions, and adjust if necessary.

jmarshrossney:
Small changes to the structure, hopefully in line with what was meant by
> ```
> loaded_model, action, batch_size
> |_ chain_autocorrelation
> |  |
> |__|_ sample
> ```

Currently three different 'batch sizes' are hard-coded:
1. The thermalisation time
2. The number of configurations used for the autocorrelation estimate
3. The normal batch size

The first could be figured out automatically, but it's not urgent so I've left it for now.

I'm going to get on with uncertainties for the observables, but I'm sure there are still many improvements to be made with this pull request.

anvil/sample.py
```diff
@@ -25,6 +27,175 @@ def sample(loaded_model, action, target_length: int, n_large=20000) -> torch.Ten
...
+        phi = loaded_model.inverse_map(z)  # map using trained loaded_model to phi
+        log_ptilde = loaded_model(phi)
+    history = torch.zeros(batch_size, dtype=torch.bool)  # accept/reject history
+    chain_indices = torch.zeros(batch_size, dtype=torch.long)
+
+    log_ratio = log_ptilde + action(phi)
+    if not isfinite(exp(float(min(log_ratio) - max(log_ratio)))):
+        raise ValueError("could run into nans")
+
+    if first:  # first batch
```
wilsonmr: surely first is unneccessary when log_ratio_i and state_i are None by default, it's like extra bookkeeping for the sake of it

```python
if state_i is None and log_ratio i is None:
...
elif state_i is not None and log_ratio_i is not None:
...
else:
    raise SomeKindofError("both need to either be None or have values")
```
I mean honestly passing the log ratio is kind of extra book keeping (I know it was me that originally put that) because it means one could potentially give a pair of values which don't match. Then it simplifies massively: the function takes `state_i` then `log_ratio_i` and `first` are redundant

so then we have

```python
    with torch.no_grad():  # don't track gradients
        z = torch.randn((batch_size, loaded_model.size_in))  # random z configurations
        phi = loaded_model.inverse_map(z)  # map using trained loaded_model to phi
        if state_i is not None:
            phi[0] = state_i
        log_ptilde = loaded_model(phi)
```

jmarshrossney: Ah yeah, first is unnecessary. I think I thought it was clearer at one point but it's definitely better to only pass `state_i` around.

wilsonmr: also, state_i for the rest of this function is literally just state[j-1] and so I'm thinking we could throw a tonne of local variables away which would make the whole thing cleaner.

Finally I would get this function to return just the stack of states and the history (in that order)
then in later uses of the function we can just get initial state by indexing the final state of the returned stack

that way you won't have to have the horrible unpacking of function that you have at some point:
```python
c, d = return_abcd()[2:]
```

jmarshrossney: I do agree with the neatness side of things, but is there not an efficiency argument for avoiding indexing arrays more times than necessary?

anvil/sample.py
```diff
@@ -25,6 +27,175 @@ def sample(loaded_model, action, target_length: int, n_large=20000) -> torch.Ten
...
+
+    tskip = ceil(2 * integrated_autocorrelation)
+    print(
+        f"Guess for sampling interval: {tskip}, based on {batch_size} configurations."
+    )
+
+    return state, log_ratio, tskip
+
+
+def sample(loaded_model, action, target_length: int) -> torch.Tensor:
```
wilsonmr: ok so this is more of a reportengine thing:
rename `thermalisation` -> `thermalised_state` and have it just return a single state

then make the signature of `chain_autocorrelation`: 

```python
def chain_autocorrelation(loaded_model, action, thermalised_state):
```

then signature of `sample`:

```python
def sample(loaded_model, action, target_length: int, thermalised_state, chain_autocorrelation)
```
and you can skip the first bit of the function because report engine will automatically evaluate `thermalised_state` and `chain_autocorrelation`

the nice part of this is that both of those functions will also be valid `actions` and so if I wanted to I could calculate auto correlation without having to do a full sample

jmarshrossney: that's rather cool

anvil/sample.py
wilsonmr: well yeah there would be except we have to index twice to save the state anyway:

```python
import torch
import numpy as np

BIG_NUM = 100000
def test_1():
    a = torch.rand(BIG_NUM)
    b = torch.zeros(BIG_NUM)
    for i in range(1, BIG_NUM):
        if np.random.rand() > 0.5:
            b[i] = a[i]
        else:
            b[i] = a[i-1]
    return b

def test_2():
    a = torch.rand(BIG_NUM)
    last = a[0]
    b = torch.zeros(BIG_NUM)
    for i in range(1, BIG_NUM):
        if np.random.rand() > 0.5:
            b[i] = a[i]
            last = a[i]
        else:
            b[i] = last
    return b

In [1]: %timeit test_1()                                                       
765 ms ± 74.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [2]: %timeit test_2()                                                       
922 ms ± 8.89 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

```

in fact we see that the second is slower

anvil/sample.py
```diff
@@ -25,6 +26,165 @@ def sample(loaded_model, action, target_length: int, n_large=20000) -> torch.Ten
...
+    Returns
+    -------
+    state: torch.Tensor
+        the current phi state, for continuity between batches
+    log_ratio: float
+        the ratio associated with state
+    """
+    t_therm = 1000 # ideally come up with a way of working this out on the fly
+
+    state = sample_batch(
```
wilsonmr: if not using values from an unpack I think it's cleaner to put

`_, _, state = sample_batch(..)`

if somebody reads your code they start at the beginning of the line and can see here that sample batch returns a tuple of 3 things and you only use the final thing

obvs if you do get it to return 2 things then it'd be more like

```
states, _ = sample_batch(...)
return states[-1, :]
```

jmarshrossney: thanks - that's a very useful tip

anvil/sample.py
jmarshrossney: I couldn't get it to work though. I always seem to run into an error if I try to use these reportengine actions as normal functions to which I can give arguments which aren't other actions.

Serious question: am I being incredibly thick?

wilsonmr: urm can you approximately write down signatures of functions that you tried

so just the `def function(arg1, arg2)` etc. and their actual names for relevant functions

wilsonmr: for me this works:

```python
def thermalised_state(loaded_model, action):
    ...
    return states[-1]


def chain_autocorrelation(loaded_model, action, thermalised_state) -> float:
    ...
    # Sample some states
    states, history = sample_batch(
            loaded_model, action, batch_size, thermalised_state
    )
    ...
    integrated_autocorrelation = 0.5 + torch.sum(
        autocorrelations / (N - torch.arange(N - 1, dtype=torch.float))
    ) # here you had a mismatch of types
    ...
    return states[-1], sample_interval


def sample(loaded_model, action, target_length: int, thermalised_state, chain_autocorrelation) -> torch.Tensor:
    ...
    current_state = thermalised_state
    _, sample_interval = chain_autocorrelation
    ...
    return decorrelated_chain
```

wilsonmr: since chain_autocorrelation is also returning a final state I could even remove thermalised state from sample and just do `current_state, sample_interval = chain_autocorrelation`

however I really would be tempted to make chain_autocorrelation return just a number - we don't really care about thermalising further we just want to know the autocorrelation

anvil/sample.py
```diff
@@ -25,6 +26,158 @@ def sample(loaded_model, action, target_length: int, n_large=20000) -> torch.Ten
...
+            consecutive_rejections = 0
+        else:  # move rejected
+            consecutive_rejections += 1
+    if consecutive_rejections > 0:  # pick up last rejection run
+        autocorrelations[1 : consecutive_rejections + 1] += 1
+
+    # Compute integrated autocorrelation
+    integrated_autocorrelation = 0.5 + torch.sum(
+        autocorrelations / (N - torch.arange(N - 1))
+    )
```
gfilaci: I think there are some missing counts when incrementing the autocorrelation slice, because autocorrelations[x] when x<consecutive_rejections must be incremented with a multiplicity... for example, inside consecutive_rejections=3 there are two combinations that give consecutive_rejections=2, "(0 0) 0" and "0 (0 0)".
A check for the implementation is to verify that in case of all rejections the autocorrelation function is identically 1 (see eq 16)

I don't know if I can push a new commit in this request, but a fix could be:
```ruby
    N = len(history)
    autocorrelations = torch.zeros(N, dtype=torch.float)
    consecutive_rejections = 0

    for step in history[1 : ]:
        if step == True:
            if consecutive_rejections > 0:
                autocorrelations[1 : consecutive_rejections + 1] += torch.arange(consecutive_rejections, 0, -1, dtype=torch.float)
                consecutive_rejections = 0
        else:
            consecutive_rejections += 1
    if consecutive_rejections > 0:
        autocorrelations[1 : consecutive_rejections + 1] += torch.arange(consecutive_rejections, 0, -1, dtype=torch.float)

    integrated_autocorrelation = 0.5 + torch.sum(autocorrelations / torch.arange(N, 0, -1, dtype=torch.float))
```

jmarshrossney: Good spot! Thanks!

anvil/sample.py
jmarshrossney: ah that's grand. My mistake was to try to do `current_state = thermalised_state(loaded_model, action)` etc.

wilsonmr:
This looks like it's shaping up pretty nicely

jmarshrossney:
I'd be grateful for a review of the more recent changes.

The current idea (not perfect but works) is:
1. Specify `thermalisation: 1` and `sample_interval: 1` in the runcard, run a quick simulation
2. Eye up the thermalisation and change the runcard parameter, run again
3. Change the runcard `sample_interval` to twice the displayed integrated autocorrelation time
4. Run a full simulation.

The "observer-independent" measure of integrated autocorrelation time can still be used by setting `sample_interval` to `-1` (any int < 1 will do).

Currently a few things are hard-coded:
- The time series and autocorrelation plots correspond to the volume-averaged two point function for separation (0,0). We could run through all separations and look for the one with the largest thermalisation / integrated autocorrelation, but I've not done that yet since it doesn't seem ideal.
- The two plots mentioned above are hard-coded to display the first 1000 data points, just because it makes things clearer to see. This is actually one thing that slightly annoys me about reportengine - you can't zoom in on plots :(

wilsonmr:
if you are developing and trying to work out best limits from plots, can you not throw a plt.show() in there if you want to zoom in? I never tried - I've never really used the mpl GUI

jmarshrossney:
> if you are developing and trying to work out best limits from plots, can you not throw a plt.show() in there if you want to zoom in?

That works brilliantly! You can zoom in manually and the zoomed plot will appear on the report.

wilsonmr:
when I said for development I didn't mean commiting it, I just meant a temporary thing to work out the limits or to get the results quickly

Some operating systems don't have access to TkAgg and so can't use the MPL gui which is why reportengine doesn't rely on it.

The point is to make actions which create sensible looking plots by themselves, in this case I'm guessing a small amount of logic could guess reasonable limits, or just picking a smaller target length in the runcard

jmarshrossney:
r.i.p

wilsonmr:
Yeah sorry, for instance I think TkAgg will crash on my computer. I really would advise the following workflow:

- develop actions with minimal working examples (small lattice size, short runs etc.) being tempted to try and get final results and develop a new action at the same time is probably gonna take longer because you'll inevitably have to run a few times
- if actions are expensive be sure to make a seperate table action so if the plot sucks you can load it externally and figure out the plot without rerunning

I can actually add an API to this which basically means you can use any action with the runcard style inputs so in this context you could do (if the API existed)

```
fig = API.your_custom_figure(lattice_length=.., ...)
plt.show()
```
in a seperate notebook or python script. It probably makes screwing around with plots a bit easier

jmarshrossney:
No worries, it's a good point about TkAgg.

The API sounds useful.
Although, given that I'm not too far off making a pull request for the uncertainties branch, which slightly changes how plotting occurs (just a separate file really) and where a discussion about plots is more likely to be on the cards, should we postpone the API until then?

wilsonmr:
as far as I'm concerned this is doing something which looks correct. I'm keen to merge this to facilitate you working on the uncertainties. I made a few little tweaks - largely concerning variable names and adding logs instead of just prints which really don't affect things (although it means you can actually turn off the printing now if in quiet mode I think)

looks nice

wilsonmr:
small comment but I didn't actually mean to updat the training report - the goal is eventually to have a range of these templates so that people can work out how to write their own anyway I don't care enough to make another commit

PR #22, opened 2019-11-22
Origin/uncertainties

c-j-johnston:
Made a bit of a mess with branches but only one small change in observables.py to avoid ValueError when it is raised


wilsonmr:
the diff is quite messed up so I can't tell exactly what you changed but could you explain to me why we would want to silently accept a value error here? Value error normally indicates something is wrong I'd rather be told this than it silently continue

c-j-johnston:
This is to fix an error in the arccosh for small samples/undersampled lattices when I was testing, so instead we append 0. Just realised as well there are more diffs than I thought -- I just checked the conflicts before which were essentially just the change and clearly my branches got more confused than I thought. I'll close and try to sort it!

PR #23, opened 2019-11-25
Uncertainties

jmarshrossney:
What has been done:
- Separate files for calculation of observables and for plotting & tables
- A few new plotting functions
- Bootstrap resampling for standard deviation (error bars), plus two new runcard parameters: `bootstrap_sample_size` and `bootstrap_n_samples`
- The battle between `two_point_green_function` and `two_point_function` rages on*

*I renamed a few things before `decorrelated_samples` was merged and haven't updated all of them, partly because I think this is a slightly more consistent naming system.

What still needs doing on this branch:
- [x] Plot bootstrap distribution for each observable
- [x] Tables for each plot so the data as well as the figures are saved
- [x] Replace print statements with log outputs
- [x] Massively improve upon docstrings!
- [x] Automatically choose sensible range for autocorrelation plot
- [x] Automatically choose window size for integrated autocorrelation calculation (doesn't necessarily belong on this branch but I'm currently working on it).

Issues:
- [x] At the moment the sampling is being re-done every time we want to loop over any parameter, regardless of whether it affects the sampling or not. Also, observable calculations are being done once for the plots and again for the tables. This is despite using the `collect` feature. I've clearly misunderstood how this works.

Looking further ahead:
- Add a feature which scans through `n_affine` and `hidden_nodes` and finds the smallest-valued combination at which the observables converge to satisfactory precision


jmarshrossney:
I'm not able to loop over post-processing / plotting parameters (i.e. those with nothing to do with sampling) without re-running the entire sampling phase.

For example, the parameter `window_S` in `autocorrelation_2pf` is only used in that function, but if I loop over it with
```
windows:
 - window_S: 1.5
 - window_S: 2.0

template_text: |
 {@with windows@}
 {@plot_autocorrelation_2pf@}
 {@endwith@}
```
it re-runs the sampling for the second `window_S`.
`plot_autocorrelation_2pf` only depends on `autocorrelation_2pf`.

Any advice would be appreciated!

c-j-johnston (approved):

anvil/config.py
```diff
@@ -114,6 +114,22 @@ def parse_sample_interval(self, interval: (int, type(None))):
         log.warning(f"Using user specified sample_interval: {interval}")
         return interval
 
+    def parse_bootstrap_n_samples(self, n_samples: (int, type(None))):
+        if n_samples is None:
+            return 1000  # default
```
wilsonmr: I *think* that it's better to not set defaults here and instead in the signature of the action which uses it, because then if you do something like `anvil-sample --help action` it will tell you the default value, whereas here it's completely obscured from user unless they open up the code

having said that the help messages are all messed up because reportengine can't interpret the numpy style docstring correctly..

anvil/plot.py
```diff
@@ -0,0 +1,317 @@
...
+from matplotlib.ticker import MaxNLocator
+from scipy.signal import correlate
+import torch
+from tqdm import tqdm
+
+from reportengine.table import table
+from reportengine.figure import figure
+
+
+def print_to_precision(val, err):
```
wilsonmr: what exactly is the problem with fstrings and using the general format?

i.e:
f"{val:.2g}"

wilsonmr: I think reportengine also does it's own formatting of numbers if used in tables (when tables were printed in reports) was there a problem with that?

anvil/plot.py
```diff
@@ -0,0 +1,317 @@
+from math import ceil, floor, log10, fabs
+
+import numpy as np
+import pandas as pd
```
wilsonmr: perhaps plot.py should just include plots and tables could go elsewhere? like table.py?

wilsonmr:
I'd appreciate it if you could try following this:

https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions

I appreciate some of the breaking of this is the gymnastics you're doing with collect.. I'll look into this it looks super weird

anvil/observables.py
```diff
@@ -230,65 +341,61 @@ def autocorrelation_2pf(training_geometry, volume_averaged_2pf):
...
-        x, y, r"$\tau_{\rm int} = $ %.3g" % integrated_autocorrelation, fontsize="large"
-    )
-    return fig
+
+#######################
+###     Collect     ###
+#######################
+# I was hoping that doing this would mean we could loop over plotting parameters
+# without having to re-run the sampling, but it's not achieved this aim.
+_two_point_function = collect("two_point_function", ("training_context",))
```
wilsonmr: I think you should remove pretty much all the collects here, they are already either taking things like `sample_training_output` or `training_geometry` which already use `training_context` so collecting here doesn't help at all

wilsonmr:
Ok so I have no idea what window_S is and then when I look in the config the parser is for `window` so `window_S` isn't being parsed through this. Now if you had in `config`:

```python
@element_of("windows")
def parse_window(self, S: (float, type(None))):
    return S
```

then you can have:

```
windows:
 - 1.5
 - 2.0

template_text: |
 {@with windows@}
 {@plot_autocorrelation_2pf@}
 {@endwith@}
```

and it's all a bit cleaner, I'm not sure if this will solve the sampling issue but it's a start

wilsonmr (changes requested):
Ok summarising the above:

- rename things according to convention
- get rid of unneccessary collects (these two are linked tbh^)
- don't set defaults inside config functions
- in future try using the `@element_of("<plural_of_key>")` to add `with` functionality

in addition to this:
- remove unused imports like `pandas` in `observables.py` - there are more btw
- when you add new functionality add a seperate runcard showing how to use it, eventually I'd like a small library of example runcards

anvil/plot.py
jmarshrossney: The precision being dictated by the magnitude of the uncertainty. I may be wrong but I couldn't see how to do that with fstrings etc.

anvil/plot.py
jmarshrossney: makes sense

anvil/plot.py
wilsonmr: oh I see, I think for now I just wouldn't bother formatting the numbers and then for final results worry about this, seems needlessly complicated for production

wilsonmr: like right now I don't think we're bothered if we type 1e5 +/- 1e-2 it just means we basically have no error

anvil/observables.py
jmarshrossney: Ah sure. I recently added those in thinking they were the key to being able to loop over that `window_S` parameter in autocorrelation without re-running the sampling each time, but apparently they're not.

anvil/config.py
jmarshrossney: good point. Will change.

anvil/sample.py
```diff
@@ -159,6 +159,9 @@ def chain_autocorrelation(
 
     # Sample some states
     _, history = sample_batch(loaded_model, action, batch_size, thermalised_state)
+    
+    accepted = float(torch.sum(history))
+    #print(f"Acceptance: {accepted / batch_size}")
```
wilsonmr: do we need a commented out print statement?

wilsonmr:
pretty close, I'd just like you to clean up the last few commented out print statements. I'd rather the prints in functions were logs if they are really necessary or just gone if not. I haven't checked for unneccessary imports yet but hopefully they're gone :)

anvil/sample.py
jmarshrossney: Not really. It's there because my current hacky script which runs anvil-train, anvil-sample x 10 until a given acceptance uses these print statements. I'm sure there's a better way to do this but I just needed to get something working.

I'll take them out of this branch though!

jmarshrossney:
Prints are gone. Unnecessary imports are gone.

Sampling now takes MUCH less time since the two point function calculation is no longer being repeated. This was a quick but not particularly pretty fix, so I will gladly modify if you have a strong opinion about how the TwoPointFunction class should be structured.

anvil/observables.py
```diff
@@ -3,35 +3,63 @@
...
+    def __init__(self, states, geometry, bootstrap_n_samples):
         self.geometry = geometry
-        self.sample = states
+        self.states = states
+        self.n_samples = bootstrap_n_samples
+        self.sample_size = states.size(0)
 
-    def __call__(self, x_0: int, x_1: int):
+        # Bootstrap samples of size (n_samples, sample_size, n_states)
+        self.sample_indices = np.random.choice(
```
wilsonmr: Mutating objects is discouraged in reportengine apps because the philosophy is you have nodes in your graph of dependencies (either things which are parsed/produced in config or functions defined in your providers module) and these things don't get mutated so can be used multiple times.

Now we cheated a little with this class because we instead wanted an object which could be called by other functions without having loads of boilerplate, but the the original method always created a new object in the call method and then returned it.

admittedly you set self.value = [] each time you use this method, however there is a parallel mode to anvil-sample and if one was using this then it could have unintended side effects.

anvil/observables.py
```diff
@@ -46,37 +74,67 @@ def __call__(self, x_0: int, x_1: int):
...
+        g_func_boot = torch.mean(
+            phi_shift_phi_boot_mean - phi_shift_boot_mean * phi_boot_mean, dim=1
         )
-        return g_func.mean()  # integrate over y and divide by volume
 
+        return torch.cat((g_func.view(1), g_func_boot))
 
-class VolumeAveraged2pf(GreenFunction):
     def __call__(self, x_0: int, x_1: int):
+        return self.value[x_0][x_1]
```
wilsonmr: which function is this return returning?

wilsonmr: I think we can do something better than this class, I will have a think

anvil/observables.py
```diff
@@ -85,34 +143,46 @@ def __call__(self, x_0: int, x_1: int):
...
 
-        va_2pf = (self.sample[:, shift] * self.sample).mean(dim=1) - self.sample.mean(
+        va_2pf = (self.states[:, shift] * self.states).mean(dim=1) - self.states.mean(
             dim=1
         ).pow(2)
+
         return va_2pf
 
+    def __call__(self, x_0: int, x_1: int):
+        return self.value[x_0][x_1]
```
wilsonmr: and this one

anvil/plot.py
```diff
@@ -0,0 +1,193 @@
...
+import matplotlib.pyplot as plt
+from matplotlib.ticker import MaxNLocator
+import torch
+
+from reportengine.figure import figure
+
+from anvil.observables import bootstrap
+
+@figure
+def plot_zero_momentum_2pf(zero_momentum_2pf, training_geometry):
```
wilsonmr: would be nice to have documentation in this file

wilsonmr: even if it just refers back to `zero_momentum_2pf`

anvil/plot.py
```diff
@@ -0,0 +1,193 @@
...
+        ax.set_title(title)
+
+    return fig
+
+
+@figure
+def plot_bootstrap_2pf(two_point_function):
+    x = t = 0
+    data_to_plot = two_point_function(x, t)
+    return plot_bootstrap_dist(data_to_plot, rf"$G$({x},{t})")
```
wilsonmr: I think we only need the r-string when there is a backslash in it

anvil/observables.py
```diff
@@ -3,35 +3,63 @@
...
+    c1 = torch.log1p(torch.sqrt(x * x - 1) / x)
+    return c0 + c1
 
-class GreenFunction:
-    def __init__(self, states, geometry):
+
+def bootstrap(observable):
+    """Return the standard deviation for an observable based on a number of
+    'bootstrap' samples."""
+    obs_full = observable[0]
```
wilsonmr, line 33: I don't really understand this convention, why is the 0th index something special, why not just have the boot samples [1:] as the full thing then take mean across samples and variance across samples?

anvil/plot.py
```diff
@@ -0,0 +1,193 @@
...
+    fig, ax = plt.subplots()
+    ax.set_title("Volume-averaged two point function")
+    ax.set_ylabel("$G_k(0,0)$")
+    ax.set_xlabel("$t$")
+    ax.plot(volume_averaged_2pf(0, 0), "-")
+    return fig
+
+
+@figure
+def plot_autocorrelation_2pf(autocorrelation_2pf):
```
wilsonmr: do we need 4 subplots here, why not just make 4 seperate figures? Personally I would make autocorrelation 4 seperate functions which depend on each other and just return one thing not a tuple. Then I would make a seperate plot for each one, perhaps I'm missing something though

anvil/plot.py
```diff
@@ -0,0 +1,193 @@
...
+    ax4.plot([W_opt,] * 2, [g_W.min(), g_W.max()], "r-", label=r"$W_{opt}$")
+    ax4.legend()
+
+    return fig
+
+
+def plot_bootstrap_dist(observable, label):
+    """Plot the distribution of some observable calculated using many bootstrap samples"""
+
+    def do_plot(ax, full_data, bootstrap_data, std):
```
wilsonmr: does this need to be inside the function? I didn't check super thoroughly but I don't think it's using any of the functionality of a nested function?

anvil/sample.py
```diff
@@ -71,10 +71,11 @@ def sample_batch(loaded_model, action, batch_size, current_state=None):
             "could run into nans based on minimum and maximum log of ratio of probabilities"
         )
 
+    rand_batch = uniform(size=batch_size+1)  # gen batch of random uniform numbers
     i = 0  # phi index of current state
     for j in range(1, batch_size + 1):  # j = phi index of proposed state
         condition = min(1, exp(float(log_ratio[i] - log_ratio[j])))
-        if np.random.uniform() <= condition:  # accepted
+        if rand_batch[j] <= condition:  # accepted
```
wilsonmr, line 78: is this faster than using inbuilt random?

wilsonmr:
I think it would be good to sit down and talk through these comments at some point, in particular I need to see the plots you're producing, to see whether they should be for example split up by figure and then we can make a plan on how to get this ready to merge

anvil/observables.py
jmarshrossney: At some point I was planning to add a bias-correction step which would use the observable calculated on the 'full' sample as well as the mean of the bootstrap distribution. But the plots of bootstrap distributions of observables were all fairly symmetric so I didn't think this was urgent.

I guess for now I'll tidy it up and do as you say, and we can cross the bias bridge if we ever come to it!

jmarshrossney: At some point I was planning to add a bias-correction step which would use the observable calculated on the 'full' sample as well as the mean of the bootstrap distribution. But the plots of bootstrap distributions of observables were all fairly symmetric so I didn't think this was urgent.

I guess for now I'll tidy it up and do as you say, and we can cross the bias bridge if we ever come to it!

wilsonmr:
Ok I'm fairly happy that I didn't break something here

we managed to get rid of classes in favour of functions which I think are more sensible and less risky. There are still a few things I can think of improving slightly but we can worry about this later

@marshrossney please look through and see what I changed, I would really appreciate it if in future PRs you could try and follow similar styles and avoid pitfalls like mutating classes

wilsonmr (approved):

PR #24, opened 2020-01-30
Improve training

jmarshrossney:
Implements main improvements to training:

- Remove ReLU on final **t** network layer (also structures networks into blocks of linear+activation)
- Scheduler parameters specified in runcard (patience and learning rate reduction factor being the most crucial)
- Optimizer and optimizer parameters in runcard (Adadelta / Adam currently appear to be the best options for single layer / multiple layer networks)

To do
- [ ] Allow complete specification of network in runcard
- [ ] Better way of linking optimizer runcard parameter to torch.optim functions and kwargs
- [ ] Find a way around setting the default patience in the config parser

This branch is based on uncertainties (f47da81).
I realise I'm jumping the gun with this PR, but hopefully it can be merged quickly after uncertainties because they're all minor changes.
Would it be easier to view changes if I set uncertainties as the base branch i.e. branch to be merged with?

### DIFF
--- a/anvil/observables.py
+++ b/anvil/observables.py
@@ -25,9 +25,18 @@ def arcosh(x):
     return c0 + c1
 
 
-###############################
-#####     Observables     #####
-###############################
+def bootstrap(observable):
+    """Return the standard deviation for an observable based on a number of
+    'bootstrap' samples."""
+    obs_full = observable[0]
+    obs_bootstrap = observable[1:]
+
+    variance = torch.mean((obs_bootstrap - obs_full) ** 2, axis=0)
+    #bias = torch.mean(obs_bootstrap) - obs_full  # not sure whether to use this
+
+    return variance.sqrt()
+
+
 class TwoPointFunction:
     def __init__(self, states, geometry, bootstrap_n_samples):
         self.geometry = geometry
@@ -160,7 +169,7 @@ def volume_averaged_2pf(sample_training_output, training_geometry):
     return VolumeAveraged2pf(sample_training_output, training_geometry)
 
 
-def zero_momentum_2pf(training_geometry, two_point_function, bootstrap_n_samples=100):
+def zero_momentum_2pf(training_geometry, two_point_function):
     r"""Calculate the zero momentum green function as a function of t
     \tilde{G}(t, 0) which is assumed to be in the first dimension defined as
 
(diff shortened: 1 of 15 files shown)


wilsonmr:
I'm just changed the base branch so I can see relevant changes. We will change this back once #23 is merged

wilsonmr:
on surface looks pretty good! Like you say we shoulod merge other branch first.

Personally I really dislike sectioning the files with comments, it seems really unneccessary to me and is sufficient to just group relevant dependencies together. If you are going to do it can you at least use a consistent style? Maybe it will change my mind...

to be specific:

```python
#######################################
###     Bootstrap distributions     ###
#######################################
```

```python
# --- Action --- #
```

choose one or get rid of both

wilsonmr:
looks great, can you remind me why we needed the block stuff in the network?

anvil/observables.py
```diff
@@ -14,6 +14,7 @@
 import torch
 from tqdm import tqdm
 
+
```
wilsonmr: I think we will just resolve the conflict by abandoning the changes to observable

wilsonmr:
tempted to make the runcard flat, instead of having kwarg dicts, just spec the kwargs and put them together.. I'll think about this

anvil/observables.py
wilsonmr: in fact all of the formatting changes which I guess were made by `black` we will probbaly abandon here, since otherwise the files are untouched

jmarshrossney:
> looks great, can you remind me why we needed the block stuff in the network?

No particularly good reason other than I found it easy to mess with `_block` during testing than directly with the `nn.ModuleList`s.



> tempted to make the runcard flat, instead of having kwarg dicts, just spec the kwargs and put them together.. I'll think about this

Absolutely fine by me. I was having trouble with optional parameters (optimiser and scheduler kwargs) in the parser, and specifying a dict was the neatest way I could find to get round this.

wilsonmr:
feel free to make these changes

jmarshrossney:
Changed base branch back to master after merging more recent changes, and ran the example phi^4 train and sample scripts.

Leaning towards leaving the "complete specification of network[s] in runcard" until a later PR, since they don't require tuning for the phi^4 theory.

wilsonmr:
> Leaning towards leaving the "complete specification of network[s] in runcard" until a later PR, since they don't require tuning for the phi^4 theory.

good idea! LGTM

wilsonmr (approved):
Looks fantastic, thanks for adding examples!

PR #30, opened 2020-04-02
Separate module for base distributions

jmarshrossney:
Created a module called distributions.py which currently contains just one class `NormalDist`.

This class does the following
- Handles generation of standard normal variates through the `__call__` method, which is used during both training and sampling.
- Contains method to compute the log density which is used in the loss function during training only.

The `NormDist` object is passed to `RealNVP` under the name `generator`, which I'm not entirely happy with.

It may seem redundant, but I'm finding that this system is easily extensible, since it means we don't have to have base distributions hard-coded into models.

Subsequent PR's will add another class, to distributions.py which deals with uniform distributions on N-spheres, containing methods for e.g. calculating the Jacobian for the spherical parameterisation.

### DIFF
--- a/anvil/config.py
+++ b/anvil/config.py
@@ -11,6 +11,7 @@ from reportengine.configparser import ConfigError, element_of
 from anvil.core import PhiFourAction, TrainingOutput
 from anvil.models import RealNVP
 from anvil.geometry import Geometry2D
+from anvil.distributions import NormalDist
 
 log = logging.getLogger(__name__)
 
@@ -67,8 +68,24 @@ class ConfigParser(Config):
     def parse_n_batch(self, nb: int):
         return nb
 
-    def produce_model(self, lattice_size, n_affine, network_kwargs):
-        model = RealNVP(n_affine=n_affine, size_in=lattice_size, **network_kwargs)
+    def produce_generator(
+        self,
+        n_batch: int,
+        lattice_size: int,
+        base_dist: str = "normal",
+        field_dimension: int = 1,
+    ):
+        if base_dist == "normal":
+            return NormalDist(
+                n_batch=n_batch,
+                lattice_volume=lattice_size,
+                field_dimension=field_dimension,
+            )
+        else:
+            raise NotImplementedError
+
+    def produce_model(self, generator, n_affine, network_kwargs):
+        model = RealNVP(generator=generator, n_affine=n_affine, **network_kwargs)
         return model
 
     def parse_epochs(self, epochs: int):
@@ -158,10 +175,10 @@ class ConfigParser(Config):
         log.warning(f"Using user specified sample_interval: {interval}")
         return interval
 
-    def parse_bootstrap_n_samples(self, n_samples: int):
+    def parse_n_boot(self, n_boot: int):
         if n_samples < 2:
-            raise ConfigError("bootstrap_n_samples must be greater than 1")
-        log.warning(f"Using user specified bootstrap_n_samples: {n_samples}")
+            raise ConfigError("n_boot must be greater than 1")
+        log.warning(f"Using user specified n_boot: {n_samples}")
         return n_samples
 
     @element_of("windows")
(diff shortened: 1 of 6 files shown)


anvil/distributions.py
```diff
@@ -0,0 +1,69 @@
...
+        self.n_batch = n_batch
+        self.lattice_volume = lattice_volume
+        self.field_dimension = field_dimension
+
+        # Number of components per field configuration =
+        # size of output Tensor at dimension 1
+        self.size_out = self.lattice_volume * self.field_dimension
+
+        # Pre-calculate normalisation for log density
+        self._log_normalisation = self.log_normalisation()
```
wilsonmr: probably more idiomatic to have the function start with underscore because its a "private" method and the attribute would be public

anvil/distributions.py
```diff
@@ -0,0 +1,69 @@
...
+    def log_normalisation(self) -> float:
+        """logarithm of the normalisation for the density function."""
+        print(self.size_out)
+        return log(sqrt(pow(2 * pi, self.size_out)))
+
+    def log_density(self, sample: torch.Tensor) -> torch.Tensor:
+        """Return log probability density of a sample generated from
+        the __call__ method above with default arguments!
+
+        The size of the sample (number of configurations) should be exactly
```
wilsonmr: I don't understand, log normalisation is independent of batch size and so it should be, right?

wilsonmr: in fact I think n_batch is completely unneccessary in this class?

anvil/distributions.py
jmarshrossney: ah yes, this is true.

wilsonmr:
seems like a good idea, I like it. Just look at my comments then we can merge

anvil/distributions.py
jmarshrossney: Also very true. Outdated comment.

jmarshrossney: cheers for picking up on that

jmarshrossney:
Apologies for many small commits. Assuming we can "squash" them all during the merge?

wilsonmr:
urm yeah you can do an interactive rebase (git rebase -i) you have to select a commit and then a text editor will open and you can select to either pick or squash commits and then you get a chance to edit the commit messages as well

jmarshrossney:
what about just using the "squash and merge" feature on the drop down menu next to "Merge pull request"?

wilsonmr:
oooh fancy, yeah that also works

jmarshrossney:
are you happy for me to do this or would you prefer to maintain exclusive merging privileges for the master branch?

anvil/config.py
```diff
@@ -158,10 +173,10 @@ def parse_sample_interval(self, interval: (int, type(None))):
         log.warning(f"Using user specified sample_interval: {interval}")
         return interval
 
-    def parse_bootstrap_n_samples(self, n_samples: int):
+    def parse_n_boot(self, n_boot: int):
```
wilsonmr, line 176: you rename this here but then I see that `observables.two_point_function` is still taking the old input probably needs to be consistent everywhere

wilsonmr:
So I don't mind who merges but I've added a repo so that merging is blocked unless there is approval, just to avoid any upsets. I think anyone can approve the PR so it's not really a restriction.. Also as a collaborator I think you probably can overwrite this anyway

As soon as you look at the bootstrap thing then I'm happy for this to be merged and you can handle the squash merge

anvil/config.py
jmarshrossney: Ah yes.
We currently have `n_boot` in the example runcard, though maybe that was by accident.
I'll change everything to `n_boot` unless you'd rather keep `bootstrap_n_sample` - it's probably clear enough what it means.

wilsonmr: not bothered - change to n_boot

wilsonmr (approved):

wilsonmr:
cool I'll leave the merging to you

PR #31, opened 2020-04-03
Combine inverse map and log density calculation

jmarshrossney:
Combined the inverse map from base to target distribution with the calculation of the model density.
This avoids going explicitly applying the forward map to calculate the density, as was being done before.

**Justification:**
- Speed, primarily. For the example `phi4_train.yml` runcard, it takes 1.5x longer to train for the same number of epochs when the density is calculated separately, compared to when it is merged into the inverse map. This is true for `RealNVP` for which the map and inverse are very cheap to compute, but the speed gains will be bigger for a more complicated map.
- I think this slightly simplifies the task of combining different modules. For example, we want to sandwich `RealNVP` between a projection map and its inverse, which could be easily implemented in another `torch.nn.Module` which includes a single call to `RealNVP.forward()`, rather than needing to call `inverse_map()` and `forward()` separately.

**Drawbacks:**
- Sampling is now slightly messier, with the current log density needing to be passed between batches in addition to the current state.
- Slight reduction in transparency.

### DIFF
--- a/anvil/models.py
+++ b/anvil/models.py
@@ -280,7 +280,7 @@ class RealNVP(nn.Module):
         super(RealNVP, self).__init__()
         self.generator = generator
         self.size_in = self.generator.size_out
-        
+
         self.affine_layers = nn.ModuleList(
             [
                 AffineLayer(self.size_in, affine_hidden_shape, affine_hidden_shape, i)
@@ -295,7 +295,8 @@ class RealNVP(nn.Module):
     def forward(self, z_input: torch.Tensor) -> torch.Tensor:
         r"""Function which maps simple distribution, z, to target distribution
         \phi, and at the same time calculates the density of the output
-        distribution using the change of variables formula.
+        distribution using the change of variables formula, according to 
+        eq. (8) of https://arxiv.org/pdf/1904.12072.pdf.
 
         Parameters
         ----------
--- a/anvil/models.py
+++ b/anvil/models.py
@@ -292,9 +292,10 @@ class RealNVP(nn.Module):
         """Function that maps field configuration to simple distribution"""
         raise NotImplementedError
 
-    def inverse_map(self, z_input: torch.Tensor) -> torch.Tensor:
+    def forward(self, z_input: torch.Tensor) -> torch.Tensor:
         r"""Function which maps simple distribution, z, to target distribution
-        \phi.
+        \phi, and at the same time calculates the density of the output
+        distribution using the change of variables formula.
 
         Parameters
         ----------
@@ -303,40 +304,23 @@ class RealNVP(nn.Module):
 
         Returns
         -------
-        out: torch.Tensor
+        phi_out: torch.Tensor
             stack of transformed states, which are drawn from an approximation
             of the target distribution, same shape as input.
-
+        log_density: torch.Tensor
+            logarithm of the probability density of the output distribution,
+            with shape (n_states, 1)
         """
+        # log density of base distribution
+        log_density = self.generator.log_density(z_input)
+
         phi_out = z_input
         for layer in reversed(self.affine_layers):  # reverse layers!
             phi_out = layer.inverse_coupling_layer(phi_out)
+            log_density += layer.log_det_jacobian(phi_out).view(-1, 1)
             # TODO: make this yield, then make a yield from wrapper?
-        return phi_out
 
-    def forward(self, phi_input: torch.Tensor) -> torch.Tensor:
-        r"""Returns the log of the exact probability (of model) associated with
-        each of the input states according to eq. (8) of
-        https://arxiv.org/pdf/1904.12072.pdf.
-
-        Parameters
-        ----------
-        phi_input: torch.Tensor
-            stack of input states, shape (N_states, D)
-
-        Returns
-        -------
-        out: torch.Tensor
-            column of log(\tilde p) associated with each of input states
-
-        """
-        log_jacob_contr = torch.zeros(phi_input.size()[0], 1)
-        z_out = phi_input
-        for layer in self.affine_layers:
-            log_jacob_contr += layer.log_det_jacobian(z_out).view(-1, 1)
-            z_out = layer(z_out)
-        log_simple_prob = self.generator.log_density(z_out)
-        return log_simple_prob + log_jacob_contr
+        return phi_out, log_density
 
 
 if __name__ == "__main__":
--- a/anvil/models.py
+++ b/anvil/models.py
@@ -185,7 +185,7 @@ class AffineLayer(nn.Module):
         z_b = s_out.exp() * phi_b + t_out
         return self.join_func([phi_a, z_b], dim=1)  # put back together state
 
-    def inverse_coupling_layer(self, z_input):
+    def inverse_coupling_layer(self, z_input) -> torch.Tensor:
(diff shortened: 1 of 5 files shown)


jmarshrossney:
Also fixed a typo in one of the parsers, which I missed last time, and finally changed the Metropolis algorithm to use `random.random()` to generate random numbers, which is faster than `numpy.uniform()` plus indexing.

wilsonmr:
Seems sensible, I did the double pass when I was still making sure the gradients were being tracked properly and it was all explicit but you're right it's very inefficient.

anvil/sample.py
```diff
@@ -103,9 +104,9 @@ def thermalised_state(loaded_model, action, thermalisation) -> torch.Tensor:
     if thermalisation is None:
         return thermalisation
 
-    states, _ = sample_batch(loaded_model, action, thermalisation)
+    states, current_log_density, _ = sample_batch(loaded_model, action, thermalisation)
     log.info(f"Thermalisation: discarded {thermalisation} configurations.")
-    return states[-1]
+    return states[-1], current_log_density
```
wilsonmr: ok I'm still thinking the best way to do this so don't change yet but I think that we should just make sample_batch return the full stack of both states *and* log density and then this function should index the last item for both because then it becomes canon that whenever we have a phi or a stack of phis, we also have a corresponding log_density

wilsonmr: i think the symmetry would make it easier to follow

wilsonmr:
hmm maybe this is stupid so bear with:

what if every forward-like pass returned a similar tuple, pretty sure the calls to `log_det_jacobian` are also made redundant by this

Also this could extend to generators and additional layer-like objects so I'm really looking a `tan`/`tanh` outputs here

that way whenever these things are executed sequentially we can always calc the log_density by summing the log_density contribution from all the things

pseudocode:

```python
z, simple_prob = generator()
phi, model_log_jac = model(z)
squashed_phi, squash_log_jac = tanh_layer(phi)
p_tilde = simple_prob + model_log_jac + squash_log_jac
...
```

is this dumb, what do you think?

jmarshrossney:
Actually I think that's a great idea.

It would also make it easier to stack multiple `nn.Module`s without needing to pass a parameter saying "don't include the log of the base distribution" as I've been doing up to now.

anvil/models.py
```diff
@@ -240,7 +234,17 @@ def log_det_jacobian(self, phi_input):
...
+
+    def forward(self, z_input: torch.Tensor) -> tuple:
+        """Given an stack of states z_input, returns a tuple 
+        (phi_out, log_det_jacob) containing the (inversely) transformed
+        states and the log of the jacobian determinant of the forward
+        transformation.
+        """
+        phi_out = self.inverse_coupling_layer(z_input)
+        log_det_jacob = self.log_det_jacobian(phi_out)
+        return phi_out, log_det_jacob
```
jmarshrossney: @wilsonmr is this roughly what you had in mind? (but with better docstring)

jmarshrossney:
- The base distribution `generator` objects are no longer attributes of the model (which is great!)
- Basically anything that returns a tensor of states also returns the associated log quantity, which can be the log probability density or the log Jacobian determinant from the transformation.
- Renamed inputs for the loss function in anticipation of the target density being not just the action, but including a term from a non-trivial parameterisation. Eventually I would aim to replace most calls to action by calls to `log_target_density` or something similar.

anvil/models.py
wilsonmr: the output is good yeah. Now please check this because if we get this wrong we break the whole thing lol

since phi = [phi_a, phi_b] -> inverse coupling -> [phi_a, z_b] (in other words phi_a is always left untouched by transformation)

then I think that if we had

```python
    def forward(self, z_input):
        z_a = z_input[:, self._a_ind]
        z_b = z_input[:, self._b_ind]
        s_out = self._s_forward(z_a)
        t_out = self._t_forward(z_a)
        phi_b = (z_b - t_out) * torch.exp(-s_out)
        return self.join_func([z_a, phi_b], dim=1), s_out.sum(dim=1, keepdim=True) # EDIT keepdim not keepdims
```

then we would get the same thing, with one less pass through the net.

I made a dummy_forward which was a copy paste of the above and it seems fine to me:

```python
>>> l = AffineLayer(10, (10, 10), (10, 10), 0)
>>> 
>>> import torch
>>> test_tensor = torch.rand(1, 10)
>>> 
>>> l(test_tensor)
(tensor([[0.9308, 0.7645, 0.3430, 0.7738, 0.1795, 1.0089, 0.7677, 0.5743, 0.4620,
         0.2349]], grad_fn=<CatBackward>), tensor([[0.1624]], grad_fn=<SumBackward1>))
>>> l.dummy_forward(test_tensor)
(tensor([[0.9308, 0.7645, 0.3430, 0.7738, 0.1795, 1.0089, 0.7677, 0.5743, 0.4620,
         0.2349]], grad_fn=<CatBackward>), tensor([[0.1624]], grad_fn=<SumBackward1>))
```

and we see it's faster

```python
In [1]: from anvil.models import AffineLayer                                                                                                                                                                

In [2]: import torch                                                                                                                                                                                        

In [3]: tens = torch.rand(1, 10)                                                                                                                                                                            

In [4]: l = AffineLayer(10, (10, 10), (10, 10), 0)                                                                                                                                                          

In [5]: def test_dummy(s, lay): 
   ...:     with torch.no_grad(): 
   ...:         return lay.dummy_forward(s) 
   ...:                                                                                                                                                                                                     

In [6]: def test_forward(s, lay): 
   ...:     with torch.no_grad(): 
   ...:         return lay(s) 
   ...:                                                                                                                                                                                                     

In [7]: %timeit test_dummy(tens, l)                                                                                                                                                                         
135 µs ± 359 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [8]: %timeit test_forward(tens, l)                                                                                                                                                                       
192 µs ± 848 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

and remember that speed up scales with size of the network!

jmarshrossney: ah great - more gains. I guess we're rediscovering the USP of RealNVP which is that it's dirt cheap.

I will update and run some checks.

jmarshrossney: oops @ keepdims

jmarshrossney:
Shall we keep `AffineLayer.coupling_layer` despite the fact that it is currently unused? At some point we may want to actually test that we can recover the simple base distribution from our model output.

wilsonmr:
Yeah exactly, doesn't hurt to keep it

anvil/models.py
wilsonmr: they purposely made tensors behave like numpy arrays, not sure why they changed all of the nomenclature!

wilsonmr:
Is this ready for me to give a final look over?

jmarshrossney:
It agrees with the "old" version as far as I can tell and passes the free theory tests from the adjacent PR, so I think it's good to go!

anvil/models.py
wilsonmr: @marshrossney did you change anything here? I don't see it

wilsonmr (commented):
see comment on forward of the affinelayer otherwise looks good

jmarshrossney:
I didn't actually push, my mistake. Sorry!

wilsonmr (approved):

PR #32, opened 2020-04-05
Free scalar benchmark test

jmarshrossney:
Straightforward benchmark test based on the free scalar theory.

To run the test:
- run `anvil-train runcards/train.yml`
- run `python anvil-test-script.py runcards/benchmarks.yml`

Feedback on how tests should be organised would be appreciated.

I'm also unsure of an appropriate level of automation; perhaps ideally the user would not have to separately run `anvil-train`. Perhaps there should be another installed script `anvil-benchmarks` which runs all of the tests and only reports back if they fail, rather than producing a report.

Things to fix / be added:
- [x] Fix: currently m_sq is defined separately in both train and benchmarks runcard
- [ ] Add: loop over masses. (EDIT: I don't think this is worth bothering with)
- [ ] Add: compare observables with theory. This could be done using net output fields as well as fields generated by `FreeScalarEigenmodes`, to rule out the observables code if the former fails but latter passes.


wilsonmr:
To me it looks like you created a new app to make sure that somebody doesn't accidentally use a non free field theory model with these new actions, am I correct? In which case I think now would be a great time to introduce reportengine.checks into this project, which we can walk though

Because essentially it seems like once you have a model trained to free theory, the things you are calculating here are not so different from observables, with the added bonus that we have theory predictions. Am I wrong?

wilsonmr:
when I think of tests in a code sense I think more like what we introduced with #29 

in particular these are like unit tests where we test low level functionality of code with very fast tests, deterministic outputs and compare to known results. One can set these things to run everytime somebody commits the code, this is great because when you change something provided it has tests, if the tests failed then you know you broke something. Would have been helpful with #31 for example if we want to make some performance upgrade but check we still are getting the same output. This is more difficult with code with a statistical element but there are ways around this: using seeds etc.

jmarshrossney:
> To me it looks like you created a new app to make sure that somebody doesn't accidentally use a non free field theory model with these new actions, am I correct?

That's a generous interpretation. I was just curious to see how the App thing worked and was also thinking that maybe an `anvil-run-tests` script could be cool. I wasn't proposing that as a finished product :)

> Because essentially it seems like once you have a model trained to free theory, the things you are calculating here are not so different from observables, with the added bonus that we have theory predictions. Am I wrong?

I agree. The fourier transform is obviously generic so could be in a utils module or just in observables if we end up using it. The eigenvalue result is only meaningful for the free theory. I suppose this links to the discussion of what to do when different theories have different (codes for) observables.


> when I think of tests in a code sense I think more like what we introduced with #29
> 
> in particular these are like unit tests where we test low level functionality of code with very fast tests, deterministic outputs and compare to known results. One can set these things to run everytime somebody commits the code, this is great because when you change something provided it has tests, if the tests failed then you know you broke something.

Ok I understand - that does seem very sensible.

To be fair, tests like this one are still useful even if they're not "exact". I.e. if we just eyeball the results we can check that nothing has gone seriously wrong. I've written another one for the spin models in the high-T limit and it's been super useful because it can be very obvious what's gone wrong just from glancing at a plot.

jmarshrossney:
> I think now would be a great time to introduce reportengine.checks into this project, which we can walk though

Sounds good. These are executed at compile time though? I'm also interested in how we could introduce checks which are run after a commit!

wilsonmr:
the checks run at compile time yes

I agree that this thing is useful, I will try and take a look later and rearrange slightly so push any outstanding changes

jmarshrossney:
Cheers. Let me know if you would like me to make any alterations!

As for the unticked to do's on this PR - looping over masses and adding theory predictions for observables - they're not urgent.

Actually looping over masses is probably not too useful; the thinking was that running the test at very high mass would check that the model could map the standard normal dist to a normal dist with a standard deviation *smaller* than one (which perhaps it might have struggled with if we kept a ReLU on the final s layer, meaning exp(s) >= 1). I don't think this is worth putting time into automating though.

I can add more theory predictions - free energy, two point function etc - at a later date.

wilsonmr:
I really like the idea that we can have a push button thing which outputs something we can copy and paste before merging that shows things are still working correctly. Refactoring this took a backseat last week but I will see if I can get this done this week because it looks nice

wilsonmr:
@marshrossney I've done some refactoring here, if you reinstall the code you can now run anvil-benchmark from anywhere and it will do the train/sample - it stores the output in /tmp which I think is fine..

please could you add docstrings to benchmarks.py? You might also want to change the name of some of the functions like "table_variance" to be a bit more specific

side note: in the future I think we could try seeding this whole thing and check that not only does the code get the right answer but gets the same answer everytime but I think for now its fine

anvil/checks.py
```diff
@@ -0,0 +1,20 @@
...
+
+Module containing checks. Checks are performed at "compile time" and ensure
+correct configuration is used before executing any actions
+
+"""
+
+from reportengine.checks import make_argcheck, CheckError
+
+@make_argcheck
+def check_trained_with_free_theory(training_context):
```
wilsonmr, line 12: the checks are a bit weird but basically the argcheck arguement names matter, they check the function which they decorate for matching argument and then perform checks on whatever that argument is, in this case it's the read yaml file but it could be a more complicated object.

wilsonmr: I think it will become clearer wiht additonal examples, I'll help with these on later PRs

wilsonmr:
please remove the tests directory with duplicates of everything

anvil/checks.py
jmarshrossney: seems useful!

jmarshrossney:
I've renamed free_theory.py -> free_scalar.py 

Other than that nothing has changed except docstrings. I think it's all working ok.

wilsonmr (approved):

jmarshrossney:
Although it looks like some of the sampling actions are being executed twice? This doesn't seem to happen when I run the training followed separately by the sampling..
```bash
[WARNING]: Output folder exists: /tmp/del_me_anvil_benchmark Overwriting contents
[WARNING]: Output folder exists: /tmp/del_me_anvil_benchmark Overwriting contents
[INFO]: All requirements processed and checked successfully. Executing actions.
[INFO]: All requirements processed and checked successfully. Executing actions.
[INFO]: Thermalisation: discarded 1000 configurations.
[INFO]: Thermalisation: discarded 1000 configurations.
[INFO]: Guess for sampling interval: 2, based on 10000 configurations.
[INFO]: Guess for sampling interval: 2, based on 10000 configurations.
[INFO]: Sample interval: 2
[INFO]: Sample interval: 2
```

wilsonmr:
this wasn't happening for me, one second

wilsonmr:
I think it's just the log has been duplicated because the apps initialise the logger...

jmarshrossney:
Sure - at least the actual sampling isn't being done twice.

wilsonmr:
So that's a bit hacky - we are slightly abusing the app here, I don't think it was intended to make multiple instances of these however this looks fine to me now

jmarshrossney:
I have no moral objection to abusing this app.

wilsonmr:
well me neither until something breaks and we have no idea why 🙃 anyway I've merged this - looks good for now

PR #33, opened 2020-04-05
Bootstrap function loops over batches

jmarshrossney:
Although vectorizing as much as possible is a good thing, we seemed to have slightly overstepped with the bootstrap function in that, for reasonable sample and bootstrap sample sizes, it tries to allocate more memory than is available on a typical laptop.

This is a pretty rough-and-ready fix which will do for the time being, though perhaps more sophisticated memory handling will be required in the future... gulp.

#### Timings
Here are some results for the time spent in `bootstrap_function`, using the example `phi4_sample.yml` runcard (6x6 lattice). Note that a bootstrap batch size equal to `n_boot` is just the old implementation.

| `target_length` | `n_boot` | bootstrap batch size | time spent in `bootstrap_function` |
| :---: | :---: | :---: | :---: |
| 1000 | 1000 | 1000 | 18s |
| 1000 | 1000 | 100 | 8.3s |
| 1000 | 1000 | 10 | 6.8s |
| 1000 | 1000 | 1 | 8.4s |
| 10000 | 1000 | 1000 | 10 mins (!) |
| 10000 | 1000 | 100 | 200s |
| 10000 | 1000 | 10 |  90s |
| 10000 | 1000 | 1 | 50s |

I noticed that pytorch normally uses at least 4 of my laptop's cores, but that dropped to just one for the 10 minute example which was using almost 30% of my RAM. It would be nice to understand this behaviour.

### DIFF
--- a/anvil/observables.py
+++ b/anvil/observables.py
@@ -14,7 +14,6 @@ from scipy.signal import correlate
 import torch
 
 from tqdm import tqdm
-from time import time
 
 
 def arcosh(x):
--- a/anvil/observables.py
+++ b/anvil/observables.py
@@ -13,8 +13,6 @@ import numpy as np
 from scipy.signal import correlate
 import torch
 
-from tqdm import tqdm
-
 
 def arcosh(x):
     """Inverse hyperbolic cosine function for torch.Tensor arguments.
@@ -148,21 +146,15 @@ def bootstrap_function(func, states, *args, n_boot=100):
         dimension
 
     """
-    # NOTE: Optimal seems to be around n_sample x batch_size = 10000 for
-    # calc_two_point_function on my laptop. Clearly this is not a general
-    # rule though.
     sample_size = states.shape[0]
-    batch_size = min(n_boot, max(1, 10000 // sample_size))
-    n_batch = n_boot // batch_size
 
-    pbar = tqdm(range(n_batch), desc="bootstrap batch")
     res = []
-    for n in pbar:
-        boot_index = torch.randint(0, sample_size, size=(sample_size, batch_size))
-        resampled_states = states[boot_index, :].transpose(1, 2)
+    for resample in range(n_boot):
+        boot_index = torch.randint(0, sample_size, size=(sample_size,))
+        resampled_states = states[boot_index, :]
         res.append(func(resampled_states, *args))
 
-    return torch.cat(res, dim=-1)
+    return torch.stack(res, dim=-1)
 
 
 def two_point_function(sample_training_output, training_geometry, n_boot=100):
--- a/anvil/observables.py
+++ b/anvil/observables.py
@@ -13,6 +13,9 @@ import numpy as np
 from scipy.signal import correlate
 import torch
 
+from tqdm import tqdm
+from time import time
+
 
 def arcosh(x):
     """Inverse hyperbolic cosine function for torch.Tensor arguments.
@@ -25,18 +28,6 @@ def arcosh(x):
     # return c0 + c1
     # NOTE: might need stable version here
     return torch.log(x + torch.sqrt(pow(x, 2) - 1))
-
-
-def bootstrap(observable):
-    """Return the standard deviation for an observable based on a number of
-    'bootstrap' samples."""
-    obs_full = observable[0]
-    obs_bootstrap = observable[1:]
-
-    variance = torch.mean((obs_bootstrap - obs_full) ** 2, axis=0)
-    # bias = torch.mean(obs_bootstrap) - obs_full  # not sure whether to use this
-
-    return variance.sqrt()
 
 
 def calc_two_point_function(sample_training_output, training_geometry):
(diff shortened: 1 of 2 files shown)


wilsonmr:
the 10 minute one looks odd unless you went into swap

anvil/observables.py
```diff
@@ -153,16 +148,24 @@ def bootstrap_function(func, states, *args, n_boot=100):
...
-    res = func(resampled_states, *args)
-    return res
+    # NOTE: Optimal seems to be around n_sample x batch_size = 10000 for
+    # calc_two_point_function on my laptop. Clearly this is not a general
+    # rule though.
+    sample_size = states.shape[0]
+    batch_size = min(n_boot, max(1, 10000 // sample_size))
+    n_batch = n_boot // batch_size
 
+    pbar = tqdm(range(n_batch), desc="bootstrap batch")
```
wilsonmr: I kind of hate using tqdm in sampling/observables but I guess it's useful when something takes minutes that you were expecting to take seconds, whats the longest you would expect to be in this function now?

wilsonmr: just wonder if we could `log.debug(f"bootstrap batch: {n}")` in loop so when its working fine we dont need to see this and then can run in debug mode when it takes ages to see whats happening?

anvil/observables.py
```diff
@@ -153,16 +148,24 @@ def bootstrap_function(func, states, *args, n_boot=100):
...
-    # put boot index on final dimension
-    resampled_states = states[boot_index, :].transpose(1, 2)
-    res = func(resampled_states, *args)
-    return res
+    # NOTE: Optimal seems to be around n_sample x batch_size = 10000 for
+    # calc_two_point_function on my laptop. Clearly this is not a general
+    # rule though.
+    sample_size = states.shape[0]
+    batch_size = min(n_boot, max(1, 10000 // sample_size))
+    n_batch = n_boot // batch_size
```
wilsonmr: do we actually get what we asked for with //, isn't that floor divide?

jmarshrossney:
> the 10 minute one looks odd unless you went into swap

Yeah as I said I was down to one core which I don't really understand. And the memory usage was under 30%. But I guess the point is that this can happen and since I would like to run stuff on my laptop it would be nice to avoid it!

wilsonmr:
yeah for sure. Well I think this looks fine - see what you think about my comments

anvil/observables.py
jmarshrossney: Ok, I've switched to using `ceil` so we get `n_boot` resamples at minimum. I don't think we need to be more precise than this?

wilsonmr: I should have been more explicit in asking what you think about this. If you think it's immaterial and n_boot is just approximate then I guess that's ok. I just think it's neat if when I bootstrap with n_boot = 1000 then I get 1000 resamples

anvil/observables.py
```diff
@@ -153,16 +148,24 @@ def bootstrap_function(func, states, *args, n_boot=100):
         dimension
 
     """
-    boot_index = torch.randint(0, states.shape[0], size=(states.shape[0], n_boot))
-    # put boot index on final dimension
-    resampled_states = states[boot_index, :].transpose(1, 2)
-    res = func(resampled_states, *args)
-    return res
+    # NOTE: Optimal seems to be around n_sample x batch_size = 10000 for
```
wilsonmr: surely this is also highly dependent on the lattice size?

anvil/observables.py
jmarshrossney: > I kind of hate using tqdm in sampling/observables

I see the little white bar and I feel all is right with the world.

> whats the longest you would expect to be in this function now?

Hmm it's actually not scaling too well at the moment. For target length 10000 and 1000 bootstrap samples I get 50s, 2 mins, 5 mins for L = 6, 8, 10. I will look into this.

> just wonder if we could `log.debug(f"bootstrap batch: {n}")` in loop so when its working fine we dont need to see this and then can run in debug mode when it takes ages to see whats happening?

I'm fine with that.

anvil/observables.py
jmarshrossney: True. I wasn't sure it was worth bothering with though since for any reasonably large target length and `n_boot` (e.g. 10000, 1000) the fastest bootstrap "batch size" is one even for small lattices. Could be some efficiency gain but only for cases where the target length or `n_boot` is small which are fast anyway.

jmarshrossney: but I will look into it actually

anvil/observables.py
jmarshrossney: > Hmm it's actually not scaling too well at the moment. For target length 10000 and 1000 bootstrap samples I get 50s, 2 mins, 5 mins for L = 6, 8, 10. I will look into this.

I'm being dumb. This is just because `calc_two_point_function` has much more work to do.

wilsonmr:
Honestly from your table I wonder if there is a strong argument for just abandoning the batches altogether and just loop over the bootstraps?

anvil/observables.py
wilsonmr: I was doing some back of envelope stuff looking at tensor size and basically the objects created (assuming states are floats) are (N_bootbatch * N_lattice * target_length * 4 * 1e-9) gb which for 6x6 lattice, batch size 1000 and 10000 target length is like 1.44gb - not too crazy. With 8x8 this goes up to 2.56gb and then 10x10 is 4gb

something like the two point function means we have 2 copies of tensor like this so things can actually get pretty crazy pretty fast.

jmarshrossney:
I'm inclined to agree. For target length 10000 this is basically always optimal, and if our sample is smaller it's fast enough not to worry about optimising.

wilsonmr:
Exactly!

wilsonmr (approved):

PR #34, opened 2020-04-07
More flexibility with specifying target distribution and flow model in runcard

jmarshrossney:
The O(N) non-linear sigma is basically ready for a PR. However, I've been struggling to find a satisfactory way of introducing the flexibility to specify the target distribution (action, parameterisation) and flow model (RealNVP + many others to come) in the runcard.

It may be that I have missed some key feature of reportengine's Config class, but it seems to me that it might not be designed for this level of flexibility - e.g. different actions with completely different parameters. You can do this with dictionaries but then you avoid all the features of the config parser and have to manually check that the dict has valid keys and values..

This PR currently doesn't do too much, but I thought it better to start a discussion sooner rather than later because I've already spent a fair amount of time struggling to come up with something decent.

Basically, what would be really useful going forward would be the ability to specify
- base distribution
- theory (action) + parameterisation -> target distribution
- flow model (which could be much more flexible than it is now)

in the runcard.

One idea, not implemented yet, could be to have a module `options` which contains dictionaries of the form `theories = {"phi_four": PhiFourAction}` which could be imported by `config.py`, so that string parameters from the runcard could be used to index a dict, the result which would be returned by a `produce` function. That would mean we could simplify our "manual" checks by just checking that the string input is a dictionary key.

Please let me know what you think. I think that getting this sorted will mean I can go ahead and introduce different actions, flow models etc. with much greater ease.

### DIFF
--- a/anvil/config.py
+++ b/anvil/config.py
@@ -8,13 +8,12 @@ import logging
 from reportengine.report import Config
 from reportengine.configparser import ConfigError, element_of
 
-from anvil.core import PhiFourAction, TrainingOutput
+from anvil.core import Target, TrainingOutput
 from anvil.models import RealNVP
 from anvil.geometry import Geometry2D
 from anvil.distributions import NormalDist
 
 log = logging.getLogger(__name__)
-
 
 class ConfigParser(Config):
     """Extend the reportengine Config class for anvil-specific
@@ -37,18 +36,15 @@ class ConfigParser(Config):
     def produce_geometry(self, lattice_length):
         return Geometry2D(lattice_length)
 
-    def parse_m_sq(self, m: (float, int)):
-        return m
+    def parse_theory(self, theo: str):
+        return theo
 
-    def parse_lam(self, lam: (float, int)):
-        return lam
+    def parse_theory_parameters(self, params: dict):
+        return params
 
-    def parse_use_arxiv_version(self, do_use: bool):
-        return do_use
-
-    def produce_action(self, m_sq, lam, geometry, use_arxiv_version):
-        return PhiFourAction(
-            m_sq, lam, geometry=geometry, use_arxiv_version=use_arxiv_version
+    def produce_target(self, theory, theory_parameters, geometry):
+        return Target(
+            theory=theory, theory_parameters=theory_parameters, geometry=geometry
         )
 
     def parse_hidden_nodes(self, hid_spec):
@@ -68,7 +64,7 @@ class ConfigParser(Config):
     def parse_n_batch(self, nb: int):
         return nb
 
-    def produce_generator(
+    def produce_base(
         self, lattice_size: int, base_dist: str = "normal", field_dimension: int = 1,
     ):
         if base_dist == "normal":
@@ -112,11 +108,17 @@ class ConfigParser(Config):
         with self.set_context(ns=self._curr_ns.new_child(training_output.as_input())):
             _, geometry = self.parse_from_(None, "geometry", write=False)
             _, model = self.parse_from_(None, "model", write=False)
-            _, action = self.parse_from_(None, "action", write=False)
+            _, target = self.parse_from_(None, "target", write=False)
             _, cps = self.parse_from_(None, "checkpoints", write=False)
-            _, generator = self.parse_from_(None, "generator", write=False)
+            _, base = self.parse_from_(None, "base", write=False)
 
-        return dict(geometry=geometry, model=model, action=action, checkpoints=cps, generator=generator)
+        return dict(
+            geometry=geometry,
+            model=model,
+            target=target,
+            checkpoints=cps,
+            base=base,
+        )
 
     def produce_training_geometry(self, training_output):
         with self.set_context(ns=self._curr_ns.new_child(training_output.as_input())):
--- a/anvil/config.py
+++ b/anvil/config.py
@@ -122,8 +122,8 @@ class ConfigParser(Config):
         # file - hopefully doesn't cause any issues..
         return training_output.as_input()
 
-    def produce_training_geometry(self, training_output):
-        with self.set_context(ns=self._curr_ns.new_child(training_output.as_input())):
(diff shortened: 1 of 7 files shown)


jmarshrossney:
I've also changed `generator` to `base` so we have `base`, `loaded_model` and `target` as the three contributors to the KL divergence and the Metropolis-Hastings.

jmarshrossney:
This is related to the question of how to specify some, but not all, optimizer and scheduler parameters, with the non-specified ones taking custom default values.. I never really figured out a good way to do this.

wilsonmr:
Hi @marshrossney sorry for delay I've only just got chance to look at this

So I've introduced soemthing called explicit node, hopefully this provides a template you can use later on when adding more flexibility. In short decorating a production rule (`produce_*`) with `explicit_node` means that the production rule returns a function associated with a provider name, so target is whatever action is set in the production rule, the great thibng about this is the dependency of the resulting target is whatever the dependency of the function it was set to which I think is exactly what you want here

I also changed some ConfigErrors, they already provide some functionality to suggest other options so take a look at that and think about it in future

finally I changed trianing_context to be the full training config - this is a little risky if we have conflict of names although I find this unlikely and otherwise the explicit node stuff doesn't work properly

I can get things to run with this, do you have a specific example you could test to make sure I got the signs correct with the action - oh btw I moved this to distributions because that seemed to make more sense.

anvil/config.py
```diff
@@ -37,18 +36,24 @@ def produce_lattice_size(self, lattice_length, lattice_dimension):
...
+    def parse_theory(self, theory: str):
+        return theory
+
     def parse_m_sq(self, m: (float, int)):
         return m
 
     def parse_lam(self, lam: (float, int)):
         return lam
 
-    def parse_use_arxiv_version(self, do_use: bool):
```
wilsonmr, line 46: suppose we could add this back in - it was an oversight

anvil/config.py
```diff
@@ -105,20 +115,9 @@ def produce_checkpoint(self, cp_id=None, training_output=None):
...
-        return dict(
-            geometry=geometry,
-            model=model,
-            target=target,
-            checkpoints=cps,
-            base=base,
-        )
+        # NOTE: This seems a bit hacky, exposing the entire training configuration
+        # file - hopefully doesn't cause any issues..
+        return training_output.as_input()
```
jmarshrossney, line 123: What does this do exactly?

jmarshrossney:
This is brilliant - cheers!

The sign of the action looks right. Other than just training a model to check, I could perhaps come up with some unit test to check that `target` is what we expect.

wilsonmr:
eh it's probably fine might be worth having a higher level test which checks things like this - but we can think about that later.

I was thinking and with the explicit nodes where we are selecting from some classes probably a better way to do it which will scale nicer in #35 is to have some dicts at the top like

`BASE_OPTIONS = {"<runcard_key>":  <class name>, ...} `then the functions look more like

```python
try:
    return BASE_OPTIONS[key]
except KeyError:
    raise ConfigError(f"invalid base distribution {key}", key, BASE_OPTIONS.keys())
```

rather than write a huge if else loop

this looks ok to me, shall we go for it?

anvil/config.py
wilsonmr: as_input() just returns a parsed config file of the fit (which gets copied at start of fit). the function is called as_input because this means we can use "from_" in the runcard.

So basically here instead of manually filling in the dictionary as we were before we just return the entire dictionary used to define the training settings. I think it will be ok because I don't see us having a key which means one thing in the training and a different thing in the sampling - but we will need to be careful for that (key words like batch_size springs to mind..) if we define a parse function for every keyword we use in the training and the sampling then with any luck we should spot any problems like that

Now because we return the fit config like lattice_length, lattice_dimension etc. etc. then when we use the collect function the namespace has all of the necessary dependencies to produce things like the model, the target etc. we were doing that before in the set_context block so in that sense I think this is actually nicer.

wilsonmr: does that answer your question?

jmarshrossney:
Yeah! Bin the loops

wilsonmr:
cool I think we can make this final change in #35 I'm gonna merge

anvil/config.py
jmarshrossney: Sure - I understand "parsed config file" to mean big dictionary with the stuff parsed / produced using the training / "fitting" runcard. Just couldn't find this in the reportengine doc.

So at the moment we are making this dictionary available during sampling using `training_context`. Would an alternative be to look at every parameter / action we need during sampling, and write `action: from_ training_output` in the sample runcard? I realise this is not preferable, and we couldn't use collect, but I'm just trying to understand it all better.

Also, could we not get rid of `produce_training_geometry` if everything is now contained in `training_context`?

wilsonmr: > Sure - I understand "parsed config file" to mean big dictionary with the stuff parsed / produced using the training / "fitting" runcard. Just couldn't find this in the reportengine doc.
> 
> So at the moment we are making this dictionary available during sampling using `training_context`. Would an alternative be to look at every parameter / action we need during sampling, and write `action: from_ training_output` in the sample runcard? I realise 

Not sure about action but anything which is in phi4_train.yaml like `use_arxiv_version` or whatever

`use_arxiv_version: { from_: training output }`

> this is not preferable, and we couldn't use collect, but I'm just trying to understand it all better.

well this goes down the line of having a very verbose sample config file where everything says from_: training_output but isn't necessarily bad. I feel like we always have ran a training and then want to do an analysis where we don't have to retype all the training config and so returning the full training config as a namespace via this production rule seems sensible with the caveat that we have to be careful to not double define variables

> 
> Also, could we not get rid of `produce_training_geometry` if everything is now contained in `training_context`?

possibly, I haven't really thought about it. I will check before merge

wilsonmr (approved):
I'd like to get this merged so I can rebase #32

PR #35, opened 2020-04-10
Stereographic projection + spin models

jmarshrossney:
Decided to open this PR anyway, even though I would anticipate making structural changes in PR #34 since it provides some context.

**Main additions**
- New "action" called `SpinHamiltonian` which is the classical N-vector model. This is equivalent to the action for the O(N) non-linear sigma up to a global translation.
- New `nn.Module` called `StereographicProjection` which takes an "inner flow" such as Real NVP which is a map from R^d to itself, and turns it into a map from the embedded sphere to itself using the stereographic projection method,

There is an example runcard for the training.

**Checks**
Without observables code, I've just been modifying the sample script so that it performs the action `sample_training_output` rather than report. If you uncomment the `np.savetxt` lines in `models.py` and `sample.py` during sampling, the distributions at each stage in the map can be looked at.
There is a dirty plot script in the examples folder which just needs the user to specify the lattice size and the number of angles per lattice site.

This allows some sanity checking
- XY model in the high temperature limit -> the polar angle should follow a uniform distribution.
- Heisenberg model in the high temperature limit -> polar angle should be ~ sin(theta), azimuthal uniform.
- Low temperature should see more alignment of spins -> sharper peaked distributions.


jmarshrossney:
Ok I've messed up the commit history on this one with a dodgy rebase - apologies.

I've updated this PR using `explicit_node` to provide some flexibility in parsing actions, as introduced in #34.

I've split up the `SpinHamiltonian` and `SphericalUniformDist` classes so that the N=2 (XY) and N=3 (Heisenberg) cases are separate. I also got rid of all of the functionality related to the O(N) theories for arbitrary N, because we aren't planning to study these models in the near future.

I've put the actions and Hamiltonians into a new file called `theory.py`. I understand that we don't want to have too many separate files, but to me it makes sense for these to be separated from `distributions.py`, which contains potential base distributions independent of the theory.

I still need to add docstrings to a lot of functions.

PR #37, opened 2020-04-22
Training against simple distributions

jmarshrossney:
### TL;DR: Sometimes you just want to see if a model can map a Gaussian to another Gaussian.

It's useful to be able to train models to learn simple target distributions with parameters which are easy to tune (rather than e.g. turning the quartic term off in the phi^4 theory to get a Gaussian target, or using the spin models at high temperature to get a uniform target).

We can use the same distribution classes, but it means that, while base distributions are generally used via the `__call__` method, which returns a sample and its log density, target distributions end up being used via the `log_density` method, to avoid unnecessarily generating samples each time.

I've added another action `standard_normal_distribution` alongside `normal_distribution` (which can have any mean or variance), so different Gaussians can be used as the base and the target. I think generally the reportengine dependency graph prevents using the same distribution for base and target with different parameters, without having a separate action for each. That's not a big deal though.

### Optimizer and scheduler

I've also committed further abuses with `@explicit_node`, this time with the optimizer, so that now all of the kwargs are directly accessible from the runcard. It also made setting defaults much simpler and more transparent.

I have checked to make sure that it works correctly when restarting from a checkpoint.

I realise that it made some sense to have the `loaded_optimizer` action where it was, with the other checkpoint stuff. I tried out a few other ways of organising this which kept that action, but I think this way works best, from the options I could think of.

### Issues

- Some lack of consistency regarding whether `lattice_size` or `config_size` or `geometry.length ** 2` are used for the size of the second dimension of tensors. Not urgent but it's in the back of my mind. Only really matters if we want to include a second e.g. gauge field, or go to d > 2.
- It may not be sensible to have lots of optimizer and scheduler parameters with short labels such as `lr` and `eps` (and `betas`!) given in the runcard as they are. I guess this increases the chance of an accidental name clash. They could all be given names such as `optim_lr`, `optim_eps` etc.

### Some context:
This has already thrown up some very useful results! For example, when trying to learn a Gaussian target, you can see Adam run into its "overshoot" problem as you increase the target variance, since the gradient of the "potential well" becomes too small to counter the momentum in the optimizer.

### DIFF
--- a/anvil/config.py
+++ b/anvil/config.py
@@ -9,33 +9,12 @@ from reportengine.report import Config
 from reportengine.configparser import ConfigError, element_of, explicit_node
 
 from anvil.core import TrainingOutput
-from anvil.train import (
-    adam,
-    adadelta,
-    stochastic_gradient_descent,
-    reduce_lr_on_plateau,
-)
+from anvil.train import OPTIMIZER_OPTIONS, reduce_lr_on_plateau
 from anvil.models import RealNVP
 from anvil.geometry import Geometry2D
-from anvil.distributions import *
+from anvil.distributions import BASE_OPTIONS, TARGET_OPTIONS
 
 log = logging.getLogger(__name__)
-
-BASE_OPTIONS = {
-    "standard_normal": standard_normal_distribution,
-    "normal": normal_distribution,
-    "uniform": uniform_distribution,
-    "circular_uniform": circular_uniform_distribution,
-    "von_mises": von_mises_distribution,
-    "spherical_uniform": spherical_uniform_distribution,
-}
-TARGET_OPTIONS = dict({"phi_four": phi_four_action,}, **BASE_OPTIONS)
-
-OPTIMIZER_OPTIONS = {
-    "adam": adam,
-    "adadelta": adadelta,
-    "sgd": stochastic_gradient_descent,
-}
 
 
 class ConfigParser(Config):
--- a/anvil/config.py
+++ b/anvil/config.py
@@ -9,7 +9,12 @@ from reportengine.report import Config
 from reportengine.configparser import ConfigError, element_of, explicit_node
 
 from anvil.core import TrainingOutput
-from anvil.train import adam, adadelta, stochastic_gradient_descent, reduce_lr_on_plateau
+from anvil.train import (
+    adam,
+    adadelta,
+    stochastic_gradient_descent,
+    reduce_lr_on_plateau,
+)
 from anvil.models import RealNVP
 from anvil.geometry import Geometry2D
 from anvil.distributions import *
@@ -163,11 +168,6 @@ class ConfigParser(Config):
     @explicit_node
     def produce_scheduler(self):
         return reduce_lr_on_plateau
-
-    def parse_scheduler_kwargs(self, kwargs: dict):
-        if "patience" not in kwargs:
-            kwargs["patience"] = 500  # problem setting default in config parser?
-        return kwargs
 
     def parse_target_length(self, targ: int):
         return targ
--- a/anvil/config.py
+++ b/anvil/config.py
@@ -9,11 +9,28 @@ from reportengine.report import Config
 from reportengine.configparser import ConfigError, element_of, explicit_node
 
 from anvil.core import TrainingOutput
+from anvil.train import adam, adadelta, stochastic_gradient_descent, reduce_lr_on_plateau
 from anvil.models import RealNVP
 from anvil.geometry import Geometry2D
-from anvil.distributions import normal_distribution, phi_four_action
+from anvil.distributions import *
 
 log = logging.getLogger(__name__)
+
+BASE_OPTIONS = {
+    "standard_normal": standard_normal_distribution,
+    "normal": normal_distribution,
+    "uniform": uniform_distribution,
+    "circular_uniform": circular_uniform_distribution,
+    "spherical_uniform": spherical_uniform_distribution,
+}
+TARGET_OPTIONS = dict({"phi_four": phi_four_action,}, **BASE_OPTIONS)
+
+OPTIMIZER_OPTIONS = {
+    "adam": adam,
+    "adadelta": adadelta,
+    "sgd": stochastic_gradient_descent,
+}
+
 
 class ConfigParser(Config):
     """Extend the reportengine Config class for anvil-specific
@@ -33,11 +50,17 @@ class ConfigParser(Config):
         """returns the total number of nodes on lattice"""
         return pow(lattice_length, lattice_dimension)
 
+    def produce_config_size(self, lattice_size, target_dimension=1):
+        return target_dimension * lattice_size
+
     def produce_geometry(self, lattice_length):
         return Geometry2D(lattice_length)
 
-    def parse_theory(self, theory: str):
-        return theory
+    def parse_target(self, target: str):
+        return target
+
+    def parse_base(self, base: str):
+        return base
 
     def parse_m_sq(self, m: (float, int)):
         return m
@@ -49,15 +72,26 @@ class ConfigParser(Config):
         return do_use
 
     @explicit_node
-    def produce_target(self, theory):
+    def produce_target_dist(self, target):
         """Return the function which initialises the correct action"""
-        if theory == "phi_four":
-            return phi_four_action
-        raise ConfigError(
-            f"Selected theory: {theory}, has not been implemented yet",
-            theory,
-            ["phi_four"],
-        )
+        try:
+            return TARGET_OPTIONS[target]
+        except KeyError:
+            raise ConfigError(
+                f"invalid target distribution {target}", target, TARGET_OPTIONS.keys()
+            )
+
+    @explicit_node
+    def produce_base_dist(
+        self, base,
+    ):
+        """Return the action which loads appropriate base distribution"""
+        try:
+            return BASE_OPTIONS[base]
+        except KeyError:
+            raise ConfigError(
+                f"Invalid base distribution {base}", base, BASE_OPTIONS.keys()
+            )
 
     def parse_hidden_nodes(self, hid_spec):
         return hid_spec
@@ -75,17 +109,6 @@ class ConfigParser(Config):
 
     def parse_n_batch(self, nb: int):
         return nb
-
-    @explicit_node
-    def produce_base(self, base_dist: str = "normal",):
-        """Return the action which loads appropriate base distribution"""
-        if base_dist == "normal":
-            return normal_distribution
-        raise ConfigError(
-            f"Base distribution: {base_dist}, has not been implemented yet",
-            base_dist,
-            ["normal"],
-        )
 
     def produce_model(self, lattice_size, n_affine, network_kwargs):
         model = RealNVP(size_in=lattice_size, n_affine=n_affine, **network_kwargs)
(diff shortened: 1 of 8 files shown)


jmarshrossney:
I would say this probably has a higher priority than #35 and #36, since it's a kind of "core" change, albeit very modest :)

jmarshrossney:
Updated the example training runcard `phi4_train.yml`. No dictionaries needed!

anvil/config.py
```diff
@@ -9,12 +9,35 @@
...
+    reduce_lr_on_plateau,
+)
 from anvil.models import RealNVP
 from anvil.geometry import Geometry2D
-from anvil.distributions import normal_distribution, phi_four_action
+from anvil.distributions import *
 
 log = logging.getLogger(__name__)
 
+BASE_OPTIONS = {
```
wilsonmr: so now I'm wondering if we should have base options defined distributions and then just import the dictionary

anvil/config.py
```diff
@@ -9,12 +9,35 @@
...
+from anvil.train import (
+    adam,
+    adadelta,
+    stochastic_gradient_descent,
+    reduce_lr_on_plateau,
+)
 from anvil.models import RealNVP
 from anvil.geometry import Geometry2D
-from anvil.distributions import normal_distribution, phi_four_action
+from anvil.distributions import *
```
wilsonmr: I'd really rather avoid this, if you need that many things from one module then you can just import the whole module rather than polluting the namespace.

See my other comment regarding defining the options dictionary in distribution though

anvil/config.py
```diff
@@ -9,12 +9,35 @@
...
+    "standard_normal": standard_normal_distribution,
+    "normal": normal_distribution,
+    "uniform": uniform_distribution,
+    "circular_uniform": circular_uniform_distribution,
+    "von_mises": von_mises_distribution,
+    "spherical_uniform": spherical_uniform_distribution,
+}
+TARGET_OPTIONS = dict({"phi_four": phi_four_action,}, **BASE_OPTIONS)
+
+OPTIMIZER_OPTIONS = {
```
wilsonmr: likewise, although there are fewer options here

anvil/config.py
jmarshrossney: yeah that's fair.

wilsonmr:
> I've added another action `standard_normal_distribution` alongside `normal_distribution` (which can have any mean or variance), so different Gaussians can be used as the base and the target. I think generally the reportengine dependency graph prevents using the same distribution for base and target with different parameters, without having a separate action for each. That's not a big deal though.

This is because of the way we are specifying the layers - actually with the changes you have made here we are within reaching distance of a flexible specification of the pipeline, I'm not sure if we want to make that change here/yet but anyway for now this is nice!

EDIT: by layers I mean modular parts which specify base -> some stuff -> target

> ### Optimizer and scheduler
> 
> I've also committed further abuses with `@explicit_node`, this time with the optimizer, so that now all of the kwargs are directly accessible from the runcard. It also made setting defaults much simpler and more transparent.

good idea - I think the explicit node is well suited to our needs here

> 
> I have checked to make sure that it works correctly when restarting from a checkpoint.
> 
> I realise that it made some sense to have the `loaded_optimizer` action where it was, with the other checkpoint stuff. I tried out a few other ways of organising this which kept that action, but I think this way works best, from the options I could think of.
> ### Issues
> 
>     * Some lack of consistency regarding whether `lattice_size` or `config_size` or `geometry.length ** 2` are used for the size of the second dimension o(This has also helped explain why I couldn't get the simple linear spline flow to map a uniform distribution to itself!)f tensors. Not urgent but it's in the back of my mind. Only really matters if we want to include a second e.g. gauge field, or go to d > 2.

Yeah, I guess we want to probably use lattice size right? Since `geometry.length ** 2` is fixing the dimension

> 
>     * It may not be sensible to have lots of optimizer and scheduler parameters with short labels such as `lr` and `eps` (and `betas`!) given in the runcard as they are. I guess this increases the chance of an accidental name clash. They could all be given names such as `optim_lr`, `optim_eps` etc.

short labels are nice if you're lazy but I completely agree - I'd actually prefer it if things like lr was learning_rate so that the runcard is human readable

> 
> 
> ### Some context:
> 
> This has already thrown up some very useful results! For example, when trying to learn a Gaussian target, you can see Adam run into its "overshoot" problem as you increase the target variance, since the gradient of the "potential well" becomes too small to counter the momentum in the optimizer.

jmarshrossney:
> short labels are nice if you're lazy but I completely agree - I'd actually prefer it if things like lr was learning_rate so that the runcard is human readable

I'm just thinking it may be easier in practice to have a consistent rule with optimizer parameter names, e.g. "parameter" becomes "optim_parameter" in the runcard.
Otherwise the user has to first look up the available parameters for the optimizer in the PyTorch docs, then look up what the corresponding runcard parameter is by scrolling through `train.py`.

That said I'm not massively fussed about this, and happy to go ahead and rename things like "learning_rate", "optimizer_stability_factor" etc.

wilsonmr:
> I'm just thinking it may be easier in practice to have a consistent rule with optimizer parameter names, e.g. "parameter" becomes "optim_parameter" in the runcard.
> Otherwise the user has to first look up the available parameters for the optimizer in the PyTorch docs, then look up what the corresponding runcard parameter is by scrolling through `train.py`.

actually the reportengine exposes any runcard options via the help:

```
$ anvil-train --help
usage: anvil-train [-q | -d] [--style STYLE] [--formats FORMATS [FORMATS ...]]
                   [-x EXTRA_PROVIDERS [EXTRA_PROVIDERS ...]]
                   [--parallel | --no-parrallel] [-h [HELP]]
                   [-o OUTPUT | -r RETRAIN]
                   config_yml

The installed provider modules are:

 - anvil.train
 - anvil.checkpoint

Use anvil-train --help <provider module> to get specific information about actions
in the module.

Use anvil-train --help <action> to get specific information about the action.

...
```

```
$ anvil-train --help anvil.train
anvil.train
train.py

Module containing functions required to train model
The following providers are defined in this module:

    adadelta: 

    adam: 

    reduce_lr_on_plateau: 

    rms_prop: 

    shifted_kl: Sample mean of the shifted Kullbach-Leibler divergence
  between target and model distribution.
  
  Parameters ---------- model_log_density: torch.Tensor
      column of log (\tilde p) for a sample of states, which is
  returned by
      forward pass of `RealNVP` model target_log_density:
  torch.Tensor
      column of log(p) for a sample of states, which includes the
  negative
      action -S(\phi) and a possible contribution to the volume
  element due
      to a non-trivial parameterisation.
  
  Returns ------- out: torch.Tensor
      torch tensor with single element, corresponding to the
  estimation of the
      shifted K-L for set of sample states.

    stochastic_gradient_descent: 

    train: training loop of model
```

```
$ anvil-train --help adam
adam

Defined in: anvil.train

adam(loaded_model, loaded_checkpoint, *, lr=0.001, betas=(0.9,
  0.999), eps=1e-08, weight_decay=0, amsgrad=False)

The following resources are read from the configuration:

    cp_id(int or NoneType): 
  [Used by checkpoint]

    training_output(str): 
  [Used by checkpoint]

    lattice_length(int): 
  [Used by lattice_size]

    lattice_dimension(int): Parse lattice dimension from runcard
  [Used by lattice_size]

    n_affine(int): 
  [Used by model]

    hidden_nodes: 
  [Used by network_kwargs]

The following additionl arguments can be used to control the
behaviour. They are set by default to sensible values:

  lr = 0.001
  betas = (0.9, 0.999)
  eps = 1e-08
  weight_decay = 0
  amsgrad = False

```

wilsonmr:
Honestly looks fine. I was wondering if in the end we might want to use lists of namespaces to construct both the model and the network

```yaml

model_specs:
 - name: "standard_normal"
   label: "base distribution" # this has no effect but we might want it idk
   <param key>: ...
 - name: "nvp" # next 'layer' in the model is the model itself
   network_specs:
    - n_out: 10
       activation: relu
    ...
    - activation: linear
 ...
 - name: "phi_four" # final layer in model is the target
    <param key>: ...

```

then essentially we use a combination of checks to make sure that the initial and final layers are valid (so every base needs a generate function and every target needs a log_density) and collects to gather up everything.

I don't know if this would over complicate everything? Also perhaps we decide that the base and the action are distinct enough that they could get their own namespace. The advantage of using the namespace is it means that the target and base can be the same distribution but with different parameters

Perhaps we should look into this later. I guess the priority right now is really to have enough in the code to produce some acceptable results.

anvil/distributions.py
```diff
@@ -120,16 +299,17 @@ class PhiFourAction(nn.Module):
...
         self.lam = lam
         self.m_sq = m_sq
-        self.length = self.geometry.length
+        self.lattice_size = geometry.length ** 2
         if use_arxiv_version:
             self.version_factor = 2
         else:
             self.version_factor = 1
 
+        self.log_density = self.forward
```
wilsonmr: Are we actually using the forward method anywhere? I don't know if there are weird side effects of doing this because pytorch messes around with the forward method so that gradients are tracked.

I wonder if we can just drop the forward method all together and just have a log_density then I think (but we should check) that we could not inherit from the nn.Module class but I might be wrong

if not then I think I would prefer if we had something like `Target(nn.Module)` which had `def log_density(self, states): return self.forward(states)` and then inheritted from that instead

wilsonmr (commented):
see comment about target classes, otherwise looks good

anvil/distributions.py
jmarshrossney: > I don't know if there are weird side effects of doing this because pytorch messes around with the forward method so that gradients are tracked.

Thanks for flagging this up - I had not given this anywhere near enough thought...

I believe the only really important thing we get from subclassing `nn.Module` is that the optimizer is automatically made aware of any tunable parameters. I think we should definitely still subclass `nn.Module` for our normalising flow models because they are what we pass to the optimizer so they should contain the list of parameters.
Also, they're kind of "part of the network" and it's neat, plus we may want to add tunable parameters in future (e.g. a scaling factor inside a tanh).

However, when we do `current_loss.backward()` this really shouldn't complain if some of the operations in the tree aren't collected into a module forward method - it should (I think) be fine as long as every intermediate tensor has `requires_grad = True`.

So I suppose it makes sense to just have all of the base and target distributions as regular classes. That way everything can have a `log_density` method so it can be used as a target, and anything which can act as a base distribution can additionally have a method which does
```python
z = generate_random_sample(sample_size)
return z, self.log_density(z)
```

jmarshrossney:
> Honestly looks fine. I was wondering if in the end we might want to use lists of namespaces to construct both the model and the network
> 
> ```yaml
> model_specs:
>  - name: "standard_normal"
>    label: "base distribution" # this has no effect but we might want it idk
>    <param key>: ...
>  - name: "nvp" # next 'layer' in the model is the model itself
>    network_specs:
>     - n_out: 10
>        activation: relu
>     ...
>     - activation: linear
>  ...
>  - name: "phi_four" # final layer in model is the target
>     <param key>: ...
> ```
> 
> then essentially we use a combination of checks to make sure that the initial and final layers are valid (so every base needs a generate function and every target needs a log_density) and collects to gather up everything.
> 
> I don't know if this would over complicate everything? Also perhaps we decide that the base and the action are distinct enough that they could get their own namespace. The advantage of using the namespace is it means that the target and base can be the same distribution but with different parameters
> 
> Perhaps we should look into this later. I guess the priority right now is really to have enough in the code to produce some acceptable results.

I do like this a lot, but yeah I think as we discussed last week it's probably not the highest priority.

jmarshrossney:
@wilsonmr I think this is ready to go, do you agree?

So all of the distribution classes have a `log_density` method, and those which can act as base distributions also have a `__call__` (which could take a more informative name) that returns a random sample along with the log density.

wilsonmr (approved):

wilsonmr:
sorry I didn't see the commit! Looks good

PR #38, opened 2020-04-27
Neural networks specified in runcard

jmarshrossney:
This is a first attempt to allow for more detailed specification of the neural networks at runcard level.

So far, the things that can changed in the runcard are:
- Number and size of intermediate layers
- Activation function
- Separately, the activation function applied to the output vector
- Whether or not batch normalisation is used
- Whether or not the input vector is standardised before being passed through neural network

This is achieved by introducing an even lower-level class which is just called `NeuralNetwork`.
I have not attempted to introduce more flexibility because we haven't so far needed it.

The point of this is that some of the settings we had hard-coded for Real NVP don't work at all well for spline flows in particular. It makes sense to introduce this level of flexibility now rather than later, so testing can be made easier.

I realise that a dictionary of dictionaries in the runcard may not seem attractive, but unless I have missed something the alternative would be to have different parse and produce functions for each different class of network. This isn't great because in future we could easily have more than two distinct network classes in a flow, which we would like to tune independently.

This PR is based on #37 , and is "behind" #35 and #36. It would be great to sort this out before the latter two!

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,22 @@ python projects. To get an idea of how to do this, look at the
 `examples/train_example.py` which is discussed in
 [this section](##using-objects-in-external-code.).
 
+## Running tests
+
+If you wish to run tests you will additionally need to install pytest
+
+```bash
+conda install pytest
+```
+
+the tests can be run from any location, provided the code is installed, with
+the command
+
+```
+pytest --pyargs anvil
+```
+
+Consider adding tests whenever modifying/adding to the code.
 
 ## phi^4 example
 
--- a/anvil/benchmark_config/__init__.py
+++ b/anvil/benchmark_config/__init__.py
@@ -1,0 +1,4 @@
+import pathlib
+
+training_path = pathlib.Path(__file__).with_name('free_scalar_train.yml')
+sample_path = pathlib.Path(__file__).with_name('free_scalar_sample.yml')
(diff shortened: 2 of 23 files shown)


jmarshrossney:
By the way, sorry for changing all the `z`'s to `x`'s in `RealNVP` and `AffineLayer`. I was anticipating using z for CPN-1 elements, so used x in all of my notes. I thought it a good idea to remove any extra layers of translation between notes and code to give my brain a fighting chance of keeping track of what's going on.

jmarshrossney:
Hey @wilsonmr, are you reasonably happy with this one for the time being?

wilsonmr:
there were some duplicated functions in config which I removed. I also felt like right now we always have s network and t network so I just made those explicit keys in runcard. I then changed it so that we use collect and expand s_network_spec and t_network_spec as namespaces. This doesn't change the result of what we were doing before but it means that all of the network spec keys are checked at runtime - having said that there are so many defaults kicking around that this is largely masked

anyway provided I didn't break anything I prefer this layout. Checkpoints now also instantiates the networks, I had to do this or make models a provider module, and I preferred the former.. It doesn't feel like too much of a stretch. We could rename checkpoints if it was a real problem for you

wilsonmr (approved):
I think this looks good for now, but let me know if you don't like/understand something

jmarshrossney:
Ah yeah ok I like it, and mostly understand it.

The reason I had 's' and 't' as keys for dictionaries was that it seemed an easy way to let us declare any number of (groups of) neural networks with whatever names we wanted. That said, I suppose we might shouldn't need that many types of networks overall, so each one could be defined explicitly like 's' and 't'.

Happy to merge.

PR #45, opened 2020-06-01
fix bug in plotting known pdfs

jmarshrossney:
Fixes mistake which prevents the target pdf being plotted alongside the base and model distributions, if the target pdf is known.

### DIFF
--- a/anvil/distributions.py
+++ b/anvil/distributions.py
@@ -104,9 +104,8 @@ class UniformDist:
 
     @property
     def pdf(self):
-        x = torch.linspace(self.x_min, self.x_max, 10000)
         dens = 1 / (self.x_max - self.x_min)
-        return ((x, torch.zeros_like(x) + dens),)
+        return (([self.x_min, self.x_max], [dens, dens]),)
 
 
 class SemicircleDist:
--- a/anvil/distributions.py
+++ b/anvil/distributions.py
@@ -88,7 +88,7 @@ class UniformDist:
 
         self.x_min, self.x_max = support
 
-        self.log_density = lambda sample: torch.zeros((sample.size[0], 1))
+        self.log_density = lambda sample: torch.zeros((sample.shape[0], 1))
 
     def __call__(self, sample_size):
         """Return a tuple (sample, log_density) for a sample of 'sample_size'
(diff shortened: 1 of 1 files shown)


anvil/distributions.py
```diff
@@ -157,7 +160,7 @@ def log_density(self, sample):
     def pdf(self):
         x = torch.linspace(-self.radius, self.radius, 10000)
         dens = 2 / (pi * self.radius ** 2) * torch.sqrt(self.radius ** 2 - x ** 2)
-        return x + self.mean, dens
+        return ((x + self.mean, dens),)
```
wilsonmr, line 162: seems weird that we need to return a tuple of length one, where the first element is a tuple

wilsonmr: but I guess it's a work around for the fields only having a single dimension?

wilsonmr:
I was wondering if we can do something better for the PDF. I mean take the uniform distr for example:

```python
    @property
    def pdf(self):
        x = torch.linspace(self.x_min, self.x_max, 10000)
        dens = 1 / (self.x_max - self.x_min)
        return x, torch.zeros_like(x) + dens
```

why are we generating 10000 points to plot a straight horizontal line? Is having x and f(x) the best thing here?

I think that if PDF was just an actual function, so in this case:

```python
    def pdf(self, x):
        dens = 1 / (self.x_max - self.x_min)
        return (x >= self.x_min)*(x <= self.x_max)*dens
```
it might be neater.

The difficulty then is that the plot function would have to work out the x to feed into the function which for the higher dimension distributions would also need to be multidimensional (which is possible with linspace) but also with the dimension on the expected axis, with some kind of convention. In the end it might not make it any neater

anvil/distributions.py
jmarshrossney: Yeah that's all it is. I'm trading weirdness here for neatness in the function that plots them, but perhaps it makes more sense to have `pdf` return something sensible.

jmarshrossney:
> why are we generating 10000 points to plot a straight horizontal line? Is having x and f(x) the best thing here?

Yep it's nuts. Brain was miles behind as fingers were typing.

I think all of these `pdf` properties could be improved, potentially following the structure of scipy.stats.rv_continuous. But the only purpose they currently serve is to plot the target density when we're training against a simple distribution..

PR #47, opened 2020-06-04
New coupling layer: "non-compact projection"

jmarshrossney:
An alternative to wrapping Real NVP in projection and inverse projection maps. Instead, each coupling layer is a composition of a projection, a single affine transformation, and an inverse projection.

This is implemented as described in section 2.1.3 of https://arxiv.org/pdf/2002.02428.pdf. However, I haven't linearised the transformation close to 0 and 2pi, since so far all attempts have been computationally expensive and haven't noticeable improved the training.


jmarshrossney:
Not required for phi four

PR #50, opened 2020-06-10
Multimodal target distribution

jmarshrossney:
A class that implements target distributions that are an additive mixture ('convex combination') of simple distributions. Useful to see whether a flow is capable of mapping a unimodal distribution to a multimodal one.


jmarshrossney:
Not actually very useful.

PR #59, opened 2021-03-24
Add extra dependencies

wilsonmr:
I missed some dependencies for training, there may be more.

I added a test that would have caught this. It generates a normal distribution and checks that the mean is within 5std deviations of expected value. It's not very stringent but should basically never fail - the main purpose was to check that part of the code runs

### DIFF
--- a/conda-recipe/meta.yaml
+++ b/conda-recipe/meta.yaml
@@ -16,6 +16,8 @@ requirements:
     - reportengine
     - matplotlib
     - pytorch
+    - scipy
+    - pandas
 
 build:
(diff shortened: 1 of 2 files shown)


conda-recipe/meta.yaml
```diff
@@ -16,6 +16,8 @@ requirements:
     - reportengine
     - matplotlib
     - pytorch
+    - scipy
+    - pandas
```
wilsonmr, line 20:
```suggestion
    - scipy
    - pandas
    - numpy
```

PR #60, opened 2021-04-20
Slimmed down version that only implements phi four

jmarshrossney:
Somehow I've managed to break the training, but while I try to figure out what's happened, here's how things look in the 'slimmed down' version.

I ended up copying the production rule for `training_geometry` (not called `geometry_from_training`) for the base, target and model, so as to avoid having to use collect to run the sampling. I really don't understand how this thing works at all, but I think it must only work for production rules, not parsers; I had to change `parse_base(self, base, ...)` to `produce_base(self, latents, ...)` and use `latents` as the key in the input card. A little annoying but for the life of me I couldn't figure out how to use `parse_from_` in any other way.

Unfortunately I also had to duplicate `loaded_model` in checkpoints.py. I'm hoping this isn't necessary.

Was thinking maybe we could use `training_output` in some of the parse/production things. If it took some sort of null value when no training directory is found, then search for parameters in the input card and create a bunch of new stuff. Otherwise, search for parameters in `training_output.config`.

For testing purposes `anvil-benchmark` should run, and the input cards in benchmark_config/ are up-to-date.

Hopefully tidying this up and fixing the training can occur in parallel.

For reference, the working version of the code that's full of unused rubbish is at https://github.com/marshrossney/anvil/tree/eddie/anvil


wilsonmr:
> I really don't understand how this thing works at all, but I think it must only work for production rules, not parsers; I had to change parse_base(self, base, ...) to produce_base(self, latents, ...) and use latents as the key in the input card

`parse` functions really should be

```python
def parse_<meaningful namespace key>(self, <meaningless variable name>):
    ....
```
where `<meaningful namespace key>` MUST appear in the runcard.

If your function takes multiple inputs then it should be a production rule. But then a production rule is (schematically):

```python
def produce_<meaningful namespace key>(self, <another meaningful namespace key>, ..., < yet another meaningful namespace key>):
    ....
```

therefore the inputs to a production rule (which may be runcard keys) cannot be the same as the key which the production rules produces.

wilsonmr:
Would it not make more sense to have a version of the branch you linked which actually works with bits removed which are irrelevant and *then* make the changes to make the reportengine side of things neater?
At the moment this has a mix of introducing new model features, new plots but also fiddling with the config parser in a way that breaks things. I don't see why we can't have the first two features without the last bit

jmarshrossney:
I could try to split this in the way you suggest, though it's not the reportengine / config stuff that is causing problems with the training, it's my rearranging of other parts of the code (I noticed too late in the process that although it was running, the loss wasn't actually decreasing).

It's not ideal but I don't think it's completely necessary to fix this before we start talking about this reportengine/config stuff.

jmarshrossney:
Closing because superseded by #61 .

NB bug was due to not updating the Jacobian determinant in the batch normalisation layer. This shouldn't matter for training but breaks sampling because the model log density is wrong.

PR #61, opened 2021-04-20
Eddie

jmarshrossney:
The code as it is on Eddie. Plenty of Easter eggs.

To do:
- [x] Get rid of stuff that's not relevant for phi four
- [x] Get benchmark test working
- [x] Improve the model actions
  - [x] Remove unnecessary parameters (`sigma`, `interval`, `s_final_activation`...)
  - [x] Add learnable global rescaling layer
- [x] Tidy up training
  - [x]  optimizer and scheduler initialisation in `checkpoints.py`
- [x] Tidy up observables
  - [x] Remove fields.py
  - [x] Make sure multiprocessing still works
- [x] Tidy up neural neural network class, removing final activation function altogether
- [x] Update sampling as done in #60
  - [x] Acceptance and integrated autocorrelation time should be accessible by plot/table actions
  - Ideally replace the collect over sampling action (see #46 for why this would be nice), but only if the replacement is actually better (discuss in future)
- [x] Update layers as done in #60 but double check training still works 
- [x] Use more recent PyTorch version
  - [x] Update to use PyTorch `searchsorted`
  - [x] Update free theory test to use `torch.fft.fft2`
- Option to specify scale in `GlobalRescaling` instead of letting it be a learnable parameter, or otherwise stop it from varying after some number of iterations. Generally look at rescaling in more detail (separate PR)
- Implement layer-by-layer plotting (separate PR)
- Bootstrap the cosh fit (separate PR)
- [ ] Add docstrings
- [x] Add to set of example runcards
- Add functionality for aggregating over many models and plotting results (separate PR)


anvil/distributions.py
```diff
@@ -385,38 +444,43 @@ class PhiFourAction:
...
-        corresponding unnormalised distribution
 
-        see class Notes for details on definition of action.
-        """
+        self.__dict__.update(couplings)
+
+        if parameterisation == "standard":
+            self.nneigh_coeff = -1
+            self.quadratic_coeff = (4 + self.m_sq) / 2
+            self.quartic_coeff = self.g / 24
```
wilsonmr: I really dislike this. It obfuscates the initialisation of the class. The union of the different parameters is not so many, I would be strongly in favour of explicitly setting them (with a `get` method on the dictionary)

examples/runcards/rqs.yml
```diff
@@ -0,0 +1,33 @@
+# Lattice
+lattice_length: 6
+lattice_dimension: 2
+
+# Target
+target: phi_four
+m_sq: -4
+lam: 1.0
+use_arxiv_version: false
```
wilsonmr: this example is not up to date - do we want to keep it?

anvil/layers.py
```diff
@@ -0,0 +1,1188 @@
...
+                torch.zeros(net_out.shape[0], self.size_half, 1),
+                torch.cumsum(net_out, dim=2),
+            ),
+            dim=2,
+        )
+
+        # Sort x_b into the appropriate bin
+        # NOTE: need to make x_b contiguous, otherwise searchsorted returns nonsense
+        k_ind = searchsorted(self.x_knot_points, x_b.contiguous()) - 1
+        k_ind.unsqueeze_(dim=-1)
```
wilsonmr: I was looking at the official searchsorted and I think it really does the same thing and wouldn't require calling `contiguous`. Should we swap this now?

anvil/distributions.py
```diff
@@ -82,10 +85,13 @@ class UniformDist:
 
     def __init__(self, lattice_size, *, support):
         self.size_out = lattice_size
-
+        self.support = support
```
wilsonmr: do we use this anywhere?

anvil/layers.py
```diff
@@ -0,0 +1,1188 @@
...
+        x_a_stand = (x_a - x_a.mean()) / x_a.std()  # reduce numerical instability
+
+        t_out = self.t_network(x_a_stand)
+
+        phi_b = x_b - t_out
+
+        phi_out = self._join_func([x_a, phi_b], dim=1)
+
+        if phi_out.requires_grad is False:
+            np.savetxt(f"layer_{self.i}.txt", phi_out)
```
wilsonmr: So these get saved in the sampling loop?

anvil/layers.py
```diff
@@ -0,0 +1,1188 @@
...
+
+        if self.symmetric_networks:
+            s_out.abs_()
+
+        phi_b = (x_b - t_out) * torch.exp(-s_out)
+
+        phi_out = self._join_func([x_a, phi_b], dim=1)
+        log_density += s_out.sum(dim=1, keepdim=True)
+
+        if False:  # phi_out.requires_grad is False:
```
wilsonmr: shall we remove unreachable code?

wilsonmr (commented):
first review

anvil/layers.py
wilsonmr: ..or make it reachable?

anvil/distributions.py
jmarshrossney: Yeah it's horrible - I was too lazy to come up with something better at the time, but in #60 you have

```python
class PhiFourScalar:
    def __init__(
        self,
        geometry,
        ising_coefficient,
        quadratic_coefficient,
        quartic_coefficient,
    ):
        self.shift = geometry.get_shift()
        self.c_ising = ising_coefficient
        self.c2 = quadratic_coefficient
        self.c4 = quartic_coefficient

    @classmethod
    def from_standard(cls, geometry, *, m_sq, g):
        return cls(geometry, -1, (4 + m_sq) / 2, g / 24)

    @classmethod
    def from_bosetti2015(cls, geometry, *, beta, lam):
        return cls(geometry, -beta, 1 - 2 * lam, lam)

    @classmethod
    def from_albergo2019(cls, geometry, *, m_sq, lam):
        return cls(geometry, -2, 4 + m_sq, lam)
```
which I think is better than passing a parameterisation-dependent values to the constructor.

examples/runcards/rqs.yml
jmarshrossney: no

anvil/layers.py
jmarshrossney: yes absolutely

anvil/distributions.py
jmarshrossney: no, it's relevant if you want to try and use a uniform distribution as the base, which would be a terrible idea for phi four

anvil/layers.py
jmarshrossney: Yeah this was just a quick addition to let me plot histograms of the intermediate states. The more sensible thing would be to save the output from each pair of coupling layers inside the modified `Sequential` if the user sets some flag to True. This can definitely go.

anvil/distributions.py
wilsonmr: but I mean the actual attribute self.support doesn't appear to get used?

anvil/layers.py
jmarshrossney: remove :)

anvil/layers.py
wilsonmr: That's fine, was just checking

anvil/distributions.py
wilsonmr: Yeah I suppose the logic would have to go from the init to either a factory function or the production rule. I suppose I can live with the current format for now, but something like this in the not too distant future would make me sleep better 😉

jmarshrossney: For now I've implemented the above and added logic to the 'action' instead of production rule
```python
def phi_four(geometry, parameterisation, couplings):
    constructor = getattr(PhiFourScalar, f"from_{parameterisation}")
    return constructor(geometry, **couplings)
```

anvil/checkpoint.py
```diff
@@ -23,6 +21,39 @@ def loaded_checkpoint(checkpoint):
...
+def loaded_optimizer(
+    loaded_model,
+    loaded_checkpoint,
+    optimizer,
+    optimizer_params,
+):
+    instance = optimizer(loaded_model.parameters(), **optimizer_params)
+    if loaded_checkpoint is not None:
+        instance.load_state_dict(loaded_checkpoint["optimizer_state_dict"])
+    return instance
```
jmarshrossney: I think it's nicer to load the optimizer and scheduler here, alongside the model, and just have one action that does it rather than one per possible optimizer.

anvil/config.py
```diff
@@ -156,26 +157,47 @@ def produce_training_geometry(self, training_context):
...
+                [{"params": []}],
+                **params,
+            )
+        except TypeError as error:
+            print(error)
             raise ConfigError(
-                f"Invalid optimizer {optimizer}", optimizer, OPTIMIZER_OPTIONS.keys()
+                f"Invalid optimizer keyword argument dict. Consult documentation for {optimizer}."
             )
+        return params
```
jmarshrossney: But this I think is pointless and I regret taking the time spent writing it! It would be much easier to just not attempt to catch every error in the parser, and just let pytorch throw it's errors during runtime if it needs to. So, what I mean is, I think we should use the parser exclusively to check for things that don't make sense for the *application* (e.g. negative lattice length), and don't try to catch every bad input. If you agree, I'll change these parse functions.

wilsonmr: I do agree!

wilsonmr: if you put typing in the signature then reportengine will check for that then just check for sensible values

wilsonmr: if you get rid of the checking you can also scrap optimizer from signature right?

jmarshrossney: Yeah I'll get rid.

anvil/config.py
```diff
@@ -241,13 +175,16 @@ def parse_sample_interval(self, interval: (int, type(None))):
...
         """Size of the bootstrap sample."""
         if n_boot < 2:
-            raise ConfigError("n_boot must be greater than 1")
-        log.warning(f"Using user specified n_boot: {n_boot}")
+            raise ConfigError("bootstrap sample size must be greater than 1")
+        log.warning(f"Using user specified bootstrap sample size: {n_boot}")
         return n_boot
 
+    def produce_bootstrap_seed(self):
+        return randint(0, maxsize)
```
wilsonmr: I would prefer if there was an option to seed the seed in case you really wanted to get the same results. Having said that we can leave that as a desirable in the future.

wilsonmr (commented):
Can you delete the binary file? I assume you didn't mean to add it? 

(benchmark_config/__pycache__/...)

anvil/config.py
jmarshrossney: yeah good point. How about just "if seed not provided in runcard: return randint"?

anvil/config.py
```diff
@@ -201,3 +202,9 @@ def parse_window(self, window: float):
             raise ConfigError("window must be positive")
         log.warning(f"Using user specified window 'S': {window}")
         return window
+
+    def produce_use_multiprocessing(self):
+        """Don't use mp on MacOS"""
+        if platform.system() == "Darwin":
+            return False
+        return True
```
jmarshrossney, line 216: That's a shame. Can try to fix later on I suppose.

anvil/config.py
wilsonmr: Good idea, I'll push something

anvil/config.py
wilsonmr: It';s related to that article I sent you a while ago, the short answer is if we can't pickle it, we can't support it on Mac

wilsonmr: wrt mp that is

wilsonmr: Anyway the hack is perfectly acceptable to me right now. You can run the code on both platforms but obviously it will be faster on linux

jmarshrossney: I remember that I decided not to use the "standard" multiprocessing.Pool because it requires pickling and at that time this function needing be multiprocessed was an instance method which was complicated to pickle. Now that the correlation function is just a function it might work, but I guess it's not a priority.

wilsonmr: Nah leave it for now. I'm satisfied I can train and sample with the examples

wilsonmr (approved):
Ok this runs, and on the whole looks good, I'm happy to merge this now and then revisit some areas that we highlighted later.

jmarshrossney:
horay, sounds good to me

PR #64, opened 2021-04-26
[Do not merge] Extend post-processing and plotting to act on intermediate layers of model

jmarshrossney:
Phase 1:
- [x] Action that yields model parameters (weights and biases) layer-by-layer
- [x] Action that yields batches of configurations layer-by-layer (including base and Metropolis output)
- [x] Implement basic plotting
- [ ] Subplots to separate weights from different networks, if interesting

Phase 2:
- [x] Add runcard parameter `layer_id` to allow looping the entire analysis (including correlation functions, bootstrapping etc) over each layer in the model
- [ ] Tables + save to .csv as well as the usual plots
- [ ] Write script that reads in data from saved .csv and plots all of the layer on a single figure

Ambitions:
- Instead of just numbers, it would be nice if `layer_id` could refer to an attribute of the layers in the model. E.g., if I wanted to plot the spline layer and the MCMC output I could write `layer_ids: [spline, metropolis]`.
- Better labelling of layers in the report.

### DIFF
--- a/anvil/sample.py
+++ b/anvil/sample.py
@@ -172,7 +172,7 @@ def metropolis_hastings(
 _metropolis_hastings = collect("metropolis_hastings", ("training_context",))
 
 
-def configs(_metropolis_hastings):
+def configs_from_metropolis(_metropolis_hastings):
     return _metropolis_hastings[0][0]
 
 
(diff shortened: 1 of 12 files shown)


wilsonmr:
Just the final commit that is

jmarshrossney:
Ok done. Didn't realise that was the breaking you were referring to.

wilsonmr:
My fault for not being specific enough! If we get that merged soon then we can rebase this

anvil/sample.py
```diff
@@ -180,3 +182,30 @@ def tau_chain(_metropolis_hastings):
...
+
+    i = 1
+    for block in loaded_model:
+        v, _ = block(v, 0, negative_mag)
+        # only want coupling layers
+        if len([tensor for tensor in block.state_dict().values()]) > 1:
+            if i == layer_id:
+                return v
+            else:
+                i += 1
```
jmarshrossney, line 204: I will need to adjust this after #68 is merged, since the model is no longer necessarily a flat `Sequential` of coupling blocks.

PR #65, opened 2021-04-26
Fix CI

### LINKED ISSUE #63: Fix CI and add benchmarking
as title says

wilsonmr:
closes #63 

Obviously would be nice to do something better with benchmarking but for now I think running it is useful


wilsonmr:
I'm going to break the rules, just this once..

PR #66, opened 2021-04-26
add api to anvil

wilsonmr:
Sorry for not adding this earlier, I suspect it could have helped with experimentation.

### DIFF
--- a/anvil/api.py
+++ b/anvil/api.py
@@ -1,0 +1,75 @@
+"""
+sample_api.py
+
+This module contains the ``reportengine`` programmatic API, initialized with the
+``anvil-sample`` providers, Config and Environment.
+
+Example:
+--------
+In ``anvil/examples/runcards`` directory, run the training example:
+
+    anvil-train train.yml
+
+Then you can use the API to access anvil-sample objects in a python environment.
+For example in a python shell:
+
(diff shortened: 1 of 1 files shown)


wilsonmr:
btw @marshrossney it might be simpler to use this to write the script/notebook which samples intermediate layers and plots it. because you can load up the model using this and then loop through the layers and do the forward pass manually idk, up to you..

wilsonmr:
potentially you could use the "hack" to patch each layer output into the plotting action in lieu of the real sample

schematically:

```python
API.<plotting action>(configs=torch.zeros(....))
```

where you replace the torch tensor with the tensors you just manually created.

jmarshrossney:
Ok this looks really useful!

That said, though I'm seeing it for the first time, I'm not sure it should replace a reportengine loop over `layer_id` to sample the intermediate layers, which should work well once #64 has been properly sorted. But it's always nice to have more than one way to do something.

jmarshrossney (approved):

wilsonmr:
You're right, I think that approch is the best

but generally this should make scripts a bit easier, and then if you write functions which accept inputs from the API then it should be easier to translate a function in a script into a useable action

PR #67, opened 2021-04-27
Sphinx docs

wilsonmr:
Add some sphinx docs, they don't say much right now.

@marshrossney at some point can you try building them? Should be a case of installing sphinx, going to the `docs/sphinx` directory and running `make html`, the output should be found in `build`

If it works I will work out where to store them and we can make another workflow to build and upload them.


.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout 🛎️
+        uses: actions/checkout@v2.3.1
+      - name: Install and Build 🔧
+        run: |
+          apt-get install python3-sphinx
```
wilsonmr:
```suggestion
          sudo apt-get install -y python3-sphinx
```

wilsonmr:
I look forward to squashing this embarassing commit history haha

wilsonmr:
- [x] move readme to appropriate doc sections
- [ ] update workflow to only build and deploy docs on main branch

jmarshrossney:
works for me! This is such a good incentive to actually write decent docstrings.

wilsonmr:
Yeah, e.g: https://wilsonmr.github.io/anvil/modules/anvil/anvil.html#anvil.geometry.Geometry2D has rendered very nicely.

wilsonmr:
I think we can add latex support, might be worth it!

jmarshrossney:
That would be really nice. Especially for the coupling layer transformations which have a million indices.

wilsonmr:
There are probably mistakes but I think this is a good base!

PR #73, opened 2021-05-05
Update and add tests for distributions

### LINKED ISSUE #72: Our Gaussian class could subclass torch.distributions.Normal
We could get rid of some bulk like the normalisations etc.

wilsonmr:
closes #72 and also follows the general policy of increasing coverage of docs and tests

### DIFF
--- a/anvil/distributions.py
+++ b/anvil/distributions.py
@@ -16,14 +16,14 @@ class Gaussian(Normal):
     -------
     lattice_size: int
         Number of nodes on the lattice.
-    sigma: float
-        Standard deviation for the distribution.
-    mean: float
+    loc: float, default=0
         Mean for the distribution.
+    scale: float, default=1
+        Standard deviation for the distribution.
     """
 
-    def __init__(self, size_out, *, mean=0, sigma=1):
-        super().__init__(mean, sigma)
+    def __init__(self, size_out, *, loc=0, scale=1):
+        super().__init__(loc, scale)
         self.size_out = size_out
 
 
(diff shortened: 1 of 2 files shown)


anvil/distributions.py
```diff
@@ -24,40 +22,45 @@ class Gaussian:
         Mean for the distribution.
     """
 
-    def __init__(self, size_out, *, sigma=1, mean=0):
+    def __init__(self, size_out, *, mean=0, sigma=1):
+        super().__init__(mean, sigma)
```
jmarshrossney: Do you think we should just follow PyTorch by labelling these parameters `loc` and `scale`? This would be more consistent with the way in which we wrap around `torch.optim` (i.e. not modifying the labelling of hyperparameters)

anvil/distributions.py
```diff
@@ -82,37 +85,41 @@ class PhiFourScalar:
...
+    >>> from anvil.geometry import Geometry2D
+    >>> from anvil.distributions import PhiFourScalar
+    >>> import torch
     >>> geom = Geometry2D(2)
-    >>> action = PhiFourAction(geom, "standard", {"m_sq": 4, "g": 0})
-    >>> state = torch.rand((1, 2*2))
-    >>> action(state)
+    >>> action = PhiFourScalar.from_standard(geom, **{"m_sq": 4, "g": 0})
+    >>> state = torch.rand((1, 2*2)) # 2-D so lattice cardinality is 4
+    >>> action.log_density(state)
```
jmarshrossney: I know it's pretty anal but personally I'd replace `action` with `target` or something similar, just because `action.log_density` is a bit confusing.

jmarshrossney (approved):

jmarshrossney:
Happy to make the very minor changes suggested above, let me know what you think.

wilsonmr:
Yep go for it

jmarshrossney:
Couldn't bring myself to change `sigma` to `scale` in the config parser - it'd probably do more harm than good!

wilsonmr:
yeah the "idea" with these things is the high level interface is very well thought out and never changes (so the user can keep using their fantastic runcards) and the lower level interface can change with the wind.

In reality you always have to make these kinds of choices.

wilsonmr:
haha, oops - forgot about the tests which check the low level interface 😄

jmarshrossney:
crap..

PR #74, opened 2021-05-17
Matplotlib style sheet and visual improvements

jmarshrossney:
Added a .mplstyle file to produce plots that look consistent with the style used in our paper.

Also made some minor changes to existing figures and added a simple histogram for the magnetization. (I'll make a separate PR for bigger additions to plots - hopefully this histogram is sufficiently inoffensive that it doesn't distract from the main topic).

Currently one needs to use an extra arg `anvil-sample --style style/anvil.mplstyle` because I haven't decided/figured out whether/how to override the reportengine default (which I think is just the matplotlib default).

### DIFF
--- a/anvil/plot.py
+++ b/anvil/plot.py
@@ -9,7 +9,9 @@ module containing all actions for plotting observables
 import torch
 import numpy as np
 import matplotlib.pyplot as plt
-from matplotlib.ticker import MaxNLocator
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+from mpl_toolkits.axes_grid1.anchored_artists import AnchoredSizeBar
+from matplotlib.font_manager import FontProperties
 
 from reportengine.figure import figure, figuregen
 from reportengine import collect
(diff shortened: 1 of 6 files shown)


wilsonmr:
I think we can add a `default_style` property to the sampling app class

wilsonmr:
if you add an init file like this to `style`

```
import pathlib
defaultstyle = pathlib.Path(__file__).with_name('anvil.mplstyle')
```

then you can import it. You might need to double check that the style gets properly installed by `pip install` in a fresh env, it might require adding to the package_data in `setup.py` something like `*.mplstyle`

jmarshrossney:
works in a fresh env without modifying setup.py, but good to check.

wilsonmr:
Ah, it would appear not.. I think the behaviour is different if you use the actual install or the development install. In the latter you symlink to the git directory so the file will be present, but the point is if I were to fresh install and run `pip install .` then it wouldn't copy the file

jmarshrossney:
yeah I spoke too soon, whoops. Though I'm a little confused - the "development install" is when we do `pip install .` ..? So this is what I did and it works ok. Am I right to say the problem is with the "normal" install?

wilsonmr:
The development is when we use the `-e` option https://pip.pypa.io/en/stable/cli/pip_install/#cmdoption-e

jmarshrossney:
ah!

wilsonmr:
I appreciate both could be viewed as development installs - apologies for not being clear. I suppose I would call them binary install (through conda), manual install (pip install), development install (-e option) but that's completely my convention not universal haha

jmarshrossney:
Ok let's try this again.

No need to apologise. On a positive note, if we make the documentation clear enough for me to understand I think it would qualify as "accessible" which is probably what we want!

wilsonmr:
Yeah, I'll open an issue!

wilsonmr:
Ok, I don't understand that last error. Do we need an extra package?

jmarshrossney:
I don't either - everything passes in ~20s when I run locally, in a fresh env with `pip install .` .

jmarshrossney:
I wonder if it's the fonts. I've used the STIX family because normal text matches the style of equations. Let me remove and re-check.

wilsonmr (approved):

Issue #62, opened 2021-04-26
Parsing info from trained models could be improved.


wilsonmr:
We use `collect` in a few places where we don't need to and instead should do something like in:

https://github.com/wilsonmr/anvil/blob/dca2dfe1e7c00682b113bfa459d64ecb2d7e5778/anvil/config.py#L130

jmarshrossney:
Yeah. The only things we end up needing (so far) are:
- geometry
- base distribution
- target
- model

The model is annoying though, because if we have e.g.
```python
def produce_model_from_training(self, training_context):
```
Then (I think) we also need a separate action to load it - a clone of `loaded_model` - which is very annoying. I'm sure there must be a workaround but haven't thought of one yet.

wilsonmr:
I think that'd be fine in the sense that we would make a function which is called `loaded_trained_model` which calls `loaded_model` on the `model_from_training` and then update all sampling related things to rely on this model..

wilsonmr:
because in training we never load model from the training, do we? In sampling we never would have not trained the model

jmarshrossney:
Yeah I guess that's not too weird. I was thinking it was strange to distinguish between loading a partially trained model to continue training, and loading a (potentially partially) trained model to sample, but it's probably not a big deal.

wilsonmr:
Yeah a little weird. The loaded_model could always be replaced by an explicit node but this seems overly complicated considering the training and sampling are two different scripts.

wilsonmr:
Ok so now I need to rethink this after #68

PR #76, opened 2021-05-20
Update plots

jmarshrossney:
- Made a few changes to existing plots to make them easier to read
- Improved the fitting of cosh(...) to the correlator
  - can choose how many small-separation points to ignore in the fit
  - errors are now from a bootstrap, not the covariance of the parameter estimate

Did not:
- implement layer-wise plotting: we will probably want to discuss how best to do that, whereas I'm hoping this PR can be merged pretty quickly.

### DIFF
--- a/anvil/config.py
+++ b/anvil/config.py
@@ -185,14 +185,35 @@ class ConfigParser(Config):
         log.warning(f"Using user specified bootstrap sample size: {n_boot}")
         return n_boot
 
-    def produce_bootstrap_seed(
-        self, manual_bootstrap_seed: (int, type(None)) = None):
+    def produce_bootstrap_seed(self, manual_bootstrap_seed: (int, type(None)) = None):
         if manual_bootstrap_seed is None:
             return randint(0, maxsize)
         # numpy is actually this strict but let's keep it sensible.
-        if (manual_bootstrap_seed < 0) or (manual_bootstrap_seed > 2**32):
+        if (manual_bootstrap_seed < 0) or (manual_bootstrap_seed > 2 ** 32):
             raise ConfigError("Seed is outside of appropriate range: [0, 2 ** 32]")
         return manual_bootstrap_seed
+
+    def parse_cosh_fit_min_separation(self, n: int, training_geometry):
+        """The smallest lattice separation to include in when fitting a cosh function
+        to the correlator, so as to the extract the correlation length.
+
+        See also: ``produce_cosh_fit_window``.
+        """
+        if n > training_geometry.length // 2 - 2:
+            raise ConfigError("Not enough points to for a three-parameter fit.")
+        return n
+
+    def produce_cosh_fit_window(self, training_geometry, cosh_fit_min_separation=None):
+        """Window of values corresponding to lattice separations, within which to fit
+        a cosh function to the correlator, so as to the extract the correlation length.
+        """
+        if cosh_fit_min_separation is None:
+            return slice(1, None)  # include all but (0, 0) separation by default
+
+        return slice(
+            cosh_fit_min_separation,
+            training_geometry.length - cosh_fit_min_separation + 1,
+        )
 
     @element_of("windows")
     def parse_window(self, window: float):
--- a/anvil/observables.py
+++ b/anvil/observables.py
@@ -30,11 +30,11 @@ def fit_zero_momentum_correlator(
 
     Parameters
     ----------
-    zero_momentum_correlator: torch.Tensor
+    zero_momentum_correlator
         The two point correlation function at zero spatial momentum, i.e. the
         correlation between 1-d 'slices'.
-    training_geometry: anvil.Geometry.
-        The lattice.
+    training_geometry
+        The anvil.geometry object defining the lattice.
     cosh_fit_window: slice object
         A slice object which selects the points (i.e. separations) to include in the
         fit. In general the signal at short separations will be contaminated by
--- a/anvil/observables.py
+++ b/anvil/observables.py
@@ -14,6 +14,7 @@ import scipy.optimize as optim
 
 log = logging.getLogger(__name__)
 
+
 def cosh_shift(x, xi, A, c):
     return A * np.cosh(-x / xi) + c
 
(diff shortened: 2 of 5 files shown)


anvil/observables.py
```diff
@@ -14,38 +14,46 @@
...
+        log.warning(
+            f"Failed to fit cosh to correlation function for {failures}/{n_boot} members of the bootstrap ensemble."
         )
-        return (popt, pcov, t0)
-    except RuntimeError:
-        log.warning("Failed to fit cosh to correlation function.")
-        return None
+        if failures >= n_boot - 1:
+            log.warning("Too many failures: no fit parameters will be returned.")
+            return None
```
jmarshrossney: this is probably quite generous - if as few as two members of the bootstrap sample yield a successful fit then we just continue the analysis.

anvil/config.py
```diff
@@ -185,15 +185,36 @@ def parse_bootstrap_sample_size(self, n_boot: int):
...
+    def produce_bootstrap_seed(self, manual_bootstrap_seed: (int, type(None)) = None):
         if manual_bootstrap_seed is None:
             return randint(0, maxsize)
         # numpy is actually this strict but let's keep it sensible.
-        if (manual_bootstrap_seed < 0) or (manual_bootstrap_seed > 2**32):
+        if (manual_bootstrap_seed < 0) or (manual_bootstrap_seed > 2 ** 32):
             raise ConfigError("Seed is outside of appropriate range: [0, 2 ** 32]")
         return manual_bootstrap_seed
 
+    def parse_cosh_fit_min_separation(self, n: int, training_geometry):
```
wilsonmr, line 196: can this not just take the `lattice_length` directly?

jmarshrossney: yep, good point

anvil/observables.py
```diff
@@ -14,38 +14,46 @@
...
-    yerr = zero_momentum_correlator.std(axis=-1)
-
-    try:
-        popt, pcov = optim.curve_fit(
-            cosh_shift,
-            xdata=t[window] - T // 2,
-            ydata=y[window],
-            sigma=yerr[window],
+def fit_zero_momentum_correlator(
+    zero_momentum_correlator, training_geometry, cosh_fit_window=slice(1, None)
```
wilsonmr: I thought the point was not to need default args here?

anvil/observables.py
```diff
@@ -111,7 +119,7 @@ def magnetic_susceptibility(magnetization, abs_magnetization_squared):
 
 
 def magnetization_series(configs):
-    return configs.sum(axis=1)
+    return configs.sum(axis=1).numpy()
```
wilsonmr, line 164: would it make more sense to just cast the configs numpy earlier instead of multiple times on the observables?

wilsonmr: in other words, do we ever need `configs` to be a tensor?

examples/runcards/report.md
```diff
@@ -5,15 +5,18 @@
 {@table_two_point_scalars@}
 ## Two Point Correlator
 {@plot_two_point_correlator@}
+{@plot_two_point_correlator_error@}
```
wilsonmr, line 8: do we want to keep adding to this report, or at some point provide some more examples of individual actions or smaller reports?

anvil/observables.py
jmarshrossney: yep, indeed

anvil/observables.py
jmarshrossney: yeah it would. I mean I think it only happens the once if you discount the unused correlator calculation, but it would probably be better to cast to numpy at the end of the sampling.

anvil/config.py
jmarshrossney: I can replace `training_geometry.length` with `lattice_length` in this parse function, but not the production rule directly below. I'm sure by now I should understand this behaviour but alas, I forget.

Also, since we're moving towards removing `training_context` (which contains the entire training runcard) in favour of things more akin to `training_geometry` which just extract the specific parameters we need (#62), perhaps it does make sense to leave this as is.

examples/runcards/report.md
jmarshrossney: Yeah I have opened an issue which is almost the same as addressing this #78

anvil/observables.py
```diff
@@ -38,14 +38,13 @@ def fit_zero_momentum_correlator(
...
-        if failures >= n_boot - 1:
-            log.warning("Too many failures: no fit parameters will be returned.")
-            return None
+    n_fits = len(optimised_parameters)
+    log.info(
+        f"Cosh fit succeeded for {n_fits}/{n_boot} members of the bootstrap ensemble."
+    )
+    if n_fits < 2:
+        log.warning("Too few successful fits: no fit parameters will be returned.")
+        return None
```
jmarshrossney: I think I will remove this altogether since the user might want to not use boostrapping, in which case this will always fail.

It's probably enough that we warn the user about how many of the fits succeeded. If only 3/100 members of the bootstrap ensemble could be fit then they will know to basically ignore the fit parameters being produced.

Probably more useful to give the user the option to not plot this cosh fit as an overlay to the two point correlator.

anvil/config.py
wilsonmr: that's really weird, I don't understand that..

wilsonmr: I guess leave it for now.

anvil/observables.py
jmarshrossney: I'd rather just leave this niggle until we're messing around inside `sample.py` anyway, trying to generate configurations from each layer, if that's alright.

anvil/config.py
jmarshrossney: apologies, it doesn't work with either the parse or produce - the parse just worked because I was using the default value for `cosh_fit_min_separation` in `cosh_fit_window`..

wilsonmr: I think it's because that key is only present in the training context and so we could in theory put lattice length but we'd have to collect over the fit or write the key in the sampling runcard, idk why I didn't think about this earlier.

So definitely leave it for now and I will look into extracting parameters from the trained model in a more satisfactory way..

wilsonmr: aka the issue you linked.

wilsonmr (approved):

PR #77, opened 2021-05-20
Correct a couple of typos / missing items in docs

jmarshrossney:


jmarshrossney:
Hmm I've just read more carefully and noticed that we direct users to the `meta.yaml` file to see what they need to run tests - i.e. pytest and hypothesis at the moment. Is this clear enough? It probably is...

wilsonmr (approved):

PR #79, opened 2021-05-22
Documentation update

jmarshrossney:
Expanded docstrings, added type-checking, slightly modified sphinx conf file.

Added type hints as described in [PEP 484](https://www.python.org/dev/peps/pep-0484/) but then removed a lot of it because the type-checking built into reportengine uses entirely different conventions.


anvil/free_scalar.py
```diff
@@ -140,7 +149,7 @@ def gen_complex_normal(n_sample, sigma, real=False):
         Returns
         -------
         out: numpy.ndarray
-            complex array of shape (n_sample, *sigma.shape)
+            complex array of shape (n_sample, ``*`` sigma.shape)
```
wilsonmr: This doesn't look good in docs

```suggestion
            complex array of shape: ``(n_sample, *sigma.shape)``
```

anvil/geometry.py
```diff
@@ -5,9 +5,7 @@
 
 Module containing transformations related to geometry.
 """
-from math import ceil
-from itertools import product
-
+import itertools as it
```
wilsonmr: I don't like aliasing modules, especially here where we use it once. If you really want to import the whole module then just keep it as itertools

```suggestion
import itertools as it
```
```suggestion
import itertools
```

anvil/geometry.py
```diff
@@ -190,12 +188,16 @@ def get_shift(self, shifts: tuple = (1, 1), dims: tuple = (0, 1)) -> torch.Tenso
...
+
         Notes
         -----
         The order in which the shifts are generated is defined by the lexicographical
-        order of the Cartesian product of one-dimensional shifts. See the documentation
-        for itertools.product for details.
+        order of the Cartesian product of one-dimensional shifts. See :py:func:`itertools.product`.
         """
-        for shift_cart in product(range(self.length), range(self.length)):
+        for shift_cart in it.product(range(self.length), range(self.length)):
```
wilsonmr: corresponding change if you accept the above. Having said all that I think it was fine before.

```suggestion
        for shift_cart in itertools.product(range(self.length), range(self.length)):
```

anvil/utils.py
```diff
@@ -7,7 +7,7 @@
 from math import ceil
 import torch
 import sys
-
+import torch
```
wilsonmr, line 10: you haven't removed the import above. In fact the imports here are super messy, can you sort them out?

Standards then third party separated by a line and whilst we're here, may as well be alphabetical within those blocks

docs/sphinx/source/conf.py
```diff
@@ -12,14 +12,15 @@
 #
 import os
 import sys
-sys.path.insert(0, os.path.abspath('../../anvil/'))
+
```
wilsonmr, line 15: I typically wouldn't bother reformatting automatically generated files

anvil/checkpoint.py
```diff
@@ -9,21 +9,22 @@
 """
 from pathlib import Path
 from glob import glob
-
+from copy import deepcopy
```
wilsonmr, line 12:
```suggestion
from copy import deepcopy

```

anvil/config.py
```diff
@@ -27,34 +27,37 @@ class ConfigParser(Config):
     objects
     """
 
-    def parse_lattice_length(self, length: int):
```
wilsonmr, line 30: Can you please also fix the imports at top of this file?

standards, third party, reportengine then anvil each with their own block.

anvil/observables.py
```diff
@@ -4,17 +4,23 @@
 observables.py
 """
 import numpy as np
-from scipy.signal import correlate
-from math import ceil, pi, sin
+import scipy.signal, scipy.optimize
+import scipy.optimize as optim
```
wilsonmr: please fix this: two separate lines, no alias, remove duplication

also fix the ordering of all the imports

anvil/sample.py
```diff
@@ -9,12 +9,10 @@
 from math import exp, isfinite, ceil
 import logging
 from random import random
+import numpy as np
```
wilsonmr, line 13:
```suggestion

import numpy as np
```

wilsonmr (commented):
I find the type hints a little ugly and at some points inconsistent like in `observables`

Possibly we should get a bit more consensus before refactors like this, but nevermind my opinion isn't that strong.

There are still some issues with the imports, I think blocks separated by blank lines with standards, third party, reportengine (since it's like 2nd party) and then anvil would be my preference everywhere. In the case where there is just two imports with one standard and one third party then I don't mind if they aren't separated by a line.

anvil/free_scalar.py
jmarshrossney: > I think in general, when making a change like this which refactors a fair amount of code, it might be good to get a consensus before making the changes rather than opening a PR over the weekend with a fair amount of executive decisions. Because now I'm faced with rejecting something that was clearly a non-negligible amount of work or accepting something which I'm not a huge fan of. Anyway, my opinion isn't that strong and I don't think we want to waste time discussing this, so I'm happy to merge this once you've addressed the comments.

You're right, I apologise for that. The rationale was that I was sticking to a PEP standard and that it wouldn't matter if type hints were in the signature or the docstring since Napolean could be configured to display them either way, but after realising that this would cause issues with reportengine I just decided to continue writing docstrings rather than go back and 'fix' all of the typehints. I'm more than happy to do that and it's really not too much work.

Happy to make these changes and will do so now.

wilsonmr: > I'm more than happy to do that and it's really not too much work.

nah, I think there are better things for you to do, and this is really just a personal flavour. If nothing is broken then this is fine.

At some point it might be worth adding a few regression style tests on the observables - just to make sure they run as expected and produce similar results given the same model or something but again we can do this after submission (and perhaps I should do it to better understand some observables...)

anvil/observables.py
```diff
@@ -229,7 +229,7 @@ def magnetic_susceptibility(
 
 def magnetization_series(configs: np.ndarray) -> np.ndarray:
```
wilsonmr: I guess the configs type hint should be changed as well?

jmarshrossney:
Ok I think we're ready to merge, notwithstanding the issues around mixing PyTorch and NumPy in post-processing, and some inconsistent type hinting, which I will fix in the near future.

wilsonmr (approved):

PR #83, opened 2021-05-25
Improve training context

### LINKED ISSUE #62: Parsing info from trained models could be improved.

wilsonmr:
As discussed, I have added some `training_<key>` where key refers to a namespace key with a production rule - things like `base_dist` etc.

Since we don't actually have options for base or target I just removed the explicit node stuff there. Even if there was a choice, we don't technically need it because we could just have the logic that chose the action inside that production rule.

I then have changed the model to load from the training_context. This didn't break much apart from the tests, because when you use the API there is no training output but you may want to declare a model - in this instance there is a new function called `explicit_model` which should be called. This might mean that some examples or docs are out of date (already!) but we can worry about that after we agree this approach is what we want.

The train, report, and benchmark all seem to be working fine with this. Let me know what you think.

closes #62

### DIFF
--- a/anvil/checks.py
+++ b/anvil/checks.py
@@ -11,11 +11,11 @@ correct configuration is used before executing any actions
 from reportengine.checks import make_argcheck, CheckError
 
 @make_argcheck
-def check_trained_with_free_theory(training_context):
+def check_trained_with_free_theory(training_target_dist):
     """Check that supplied model is a free theory model which in the case of
     phi^4 means that lambda = 0
     """
-    if training_context["lam"] != 0:
+    if training_target_dist.c_quartic != 0:
         raise CheckError(
             f"Theory parameters do not correspond to free theory with lam = {lam} "
             "(should be 0)"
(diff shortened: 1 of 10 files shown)


anvil/checks.py
```diff
@@ -11,11 +11,11 @@
...
-def check_trained_with_free_theory(training_context):
+def check_trained_with_free_theory(training_target_dist):
     """Check that supplied model is a free theory model which in the case of
     phi^4 means that lambda = 0
     """
-    if training_context["lam"] != 0:
+    if training_target_dist.c_quartic != 0:
         raise CheckError(
             f"Theory parameters do not correspond to free theory with lam = {lam} "
             "(should be 0)"
```
wilsonmr: oops this is totally broken.

jmarshrossney (approved):
Sure, I think it's no bad thing to simplify things where we haven't actually implemented different options yet. I can't see the different way of constructing a model using the API becoming a problem. Fine with all of this!

